PR #10, opened 2018-12-11
[DEM-695] fix issues with measurement translation in projectq backend

quantumkoen:
This fixes the two issues present in DEM-695:

* depending on the order of _measured_id's, the result would change
* the get_probabilities method did not properly take the input register into account, and always returned the number of bits used in the circuit instead of the number of bits in the get_probabilities method


src/quantuminspire/projectq/backend_qx.py
```diff
@@ -245,54 +223,56 @@ def get_probabilities(self, qureg):
...
+            qureg = [Qubit(0), Qubit(1), Qubit(5)]
+            print(self._map_state_to_bit_string(state, qureg)  # prints '011'
+
+        """
+        mapped_state = ''
+
+        for i in range(len(qureg)):
+            logical_id = qureg[i].id
+            physical_id = self._logical_to_physical(logical_id)
+            if int(state) & (1 << physical_id):
```
qSaevar: Any need to int() and int?

src/quantuminspire/projectq/backend_qx.py
```diff
@@ -303,65 +283,105 @@ def _run(self):
...
-        try:
-            if self._verbose:
-                print('sending cqasm:')
-                print('------')
-                print(qasm)
-                print('------')
+        self._cqasm = qasm
+
+    def _execute_cqasm(self):
+        """ Execute self._cqasm, raise error if raw_text in result from API is not empty (which indicates an error)
```
qSaevar: > To avoid google docstring nitpicking, I do believe you should describe errors raised with Raises: and then list the specific error classes raised and under what condition

src/quantuminspire/projectq/backend_qx.py
```diff
@@ -112,7 +111,6 @@ def _store(self, cmd):
             print('   _allocated_qubits {0}'.format(self._allocated_qubits))
 
         if self._clear:
-            self._probabilities = dict()
```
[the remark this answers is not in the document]
quantumkoen: Not really, but added it. I don't really like this whole clear mechanics anyway, but decided to leave that as is. I did enough refactoring already :)

src/quantuminspire/projectq/backend_qx.py
quantumkoen: No, but it might also be a string (since the result from the API is decoded json, and json only allows strings as keys).

src/quantuminspire/projectq/backend_qx.py
quantumkoen: :cactus: (sorry, couldn't find the checkmark, and this was the first emoticon to show when typing colon followed by c)

PR #23, opened 2019-02-04
[DEM-738] qiskit backend change to allow re-use of classical bits

QFer:
The following piece of code is allowed now (notice the measurement in cr1[0] for both qubits):
{code}
        circ1 = QuantumCircuit(qr1, qr2, cr1, cr2)
        circ1.h(qr1[0])
        circ1.h(qr2[0])
        circ1.measure(qr1[0], cr1[0])
        circ1.h(qr2[0])
        circ1.measure(qr2[0], cr1[0])
{code}


QFer:
I updated the comment and changed dict to list

PR #24, opened 2019-02-06
[DEM-737] making memory available in Qiskit

QFer:
[DEM]-737 making memory available actually means that the ExperimentResultData of an qiskit experiment returned by the SDK backend includes the raw results from the QI simulator in the 'memory' item. The raw results are converted to the classical bit representation according to the measure statements in the circuit.
The 'memory' attribute in the backend configuration is set to True to indicate that we now support this memory data.
The jira ticket refers to pp 18 of https://qutech-sd.atlassian.net/wiki/pages/viewpageattachments.action?pageId=783810644)
I furthermore did some small changes as spelling (Qiskit is the official spelling as cQASM). Resolved some PEP8 warnings.

When there is no memory data the memory argument of ExperimentResultData may not be referenced in any way.

I refactored __convert_histogram to make use of the new function __qubit_2_classical_hex, which is also used by the new function to get the memory_data in the right format.

The unittests are adjusted with this new field. I added one unittest for the number_of_shots = 1 case, because this is a special case. The backend returns no raw_data (empty list), so no memory_data is generated.


src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -188,7 +184,13 @@ def get_experiment_results(self, qi_job):
             histogram = QuantumInspireBackend.__convert_histogram(result, measurements, result['number_of_qubits'],
                                                                   job['number_of_shots'])
             histogram_obj = Obj.from_dict(histogram)
-            experiment_result_data = ExperimentResultData(counts=histogram_obj)
+            raw_data_url = result.get('raw_data_url')
+            raw_data = self.__api.get(raw_data_url)
+            memory_data = self.__get_memory_data(raw_data, measurements, result['number_of_qubits'])
+            if memory_data:
```
qSaevar: Why not always add the `memory_data `to the `experiment_result_data`? I think it is save to assume that if there is a histogram there also is some raw data.

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -188,7 +184,13 @@ def get_experiment_results(self, qi_job):
             histogram = QuantumInspireBackend.__convert_histogram(result, measurements, result['number_of_qubits'],
                                                                   job['number_of_shots'])
             histogram_obj = Obj.from_dict(histogram)
-            experiment_result_data = ExperimentResultData(counts=histogram_obj)
+            raw_data_url = result.get('raw_data_url')
```
qSaevar: Would be nice to hide this in the api, like `api.get_raw_data(result_id)` or `api.get_raw_data_from_result(result_id)`

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -188,7 +184,13 @@ def get_experiment_results(self, qi_job):
...
+            raw_data = self.__api.get(raw_data_url)
+            memory_data = self.__get_memory_data(raw_data, measurements, result['number_of_qubits'])
```
qSaevar: Another thought, now that we are accessing the raw data, we also have the count as well. So maybe converting the histogram in line 184 is an overkill? I.e., we convert our QI histogram into a number of counts, while now there is a method to get the actual counts.

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -282,36 +284,76 @@ def _collect_measurements(experiment):
             measurements = [[index, index] for index in range(number_of_qubits)]
         return {'measurements': measurements, 'number_of_clbits': number_of_clbits}
 
+    @staticmethod
+    def __qubit_2_classical_hex(qubit_register, measurements, number_of_qubits):
```
qSaevar: Then, if there is no longer a need for a special __convert_histogram, and __get_memory data, this can be combined into one `__convert_results`?

qSaevar: or retrieve_memory_and_counts or whatever.

src/quantuminspire/qiskit/backend_qx.py
QFer: There is one exception. When number of shots = 1 there is no memory data (raw_data). In this case there is a histogram. I added a unittest for this special case.
The if is necessary because ExperimentResultData won't accept None as memory parameter.

src/quantuminspire/qiskit/backend_qx.py
QFer: I think you are right but what to do with the exception nr_shots = 1. No raw data is available from the backend. We must agree on how to construct the histogram. Issue [DEM-766]

src/quantuminspire/qiskit/backend_qx.py
QFer: I agree when the previous comment is solved

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -282,36 +284,76 @@ def _collect_measurements(experiment):
...
-            Args:
-                result (dict): The result output from the quantum inspire backend with full-
-                               state projection histogram output.
-                measurements (dict): Measured qubits/classical bits map and number of classical bits
-                number_of_qubits (int): number of qubits used in the algorithm
-                number_of_shots (int): The number of times the algorithm is executed.
+        Args:
+            result (dict): The result output from the quantum inspire backend with full-
+                           state projection histogram output.
+            measurements (dict): Measured qubits/classical bits map and number of classical bits
```
CumulonimbusCalvus:
```suggestion
            measurements (list): Measured qubits/classical bits map and number of classical bits
```

QFer: Actually what was meant here is about the field 'measurements' in the dictionary with the same name 'measurements', the field measurements must be a list (and it is).

src/quantuminspire/qiskit/backend_qx.py
qSaevar: This should be fixed in ticket DEM-766

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -274,44 +275,80 @@ def _collect_measurements(experiment):
...
-            Args:
-                result (dict): The result output from the quantum inspire backend with full-
-                               state projection histogram output.
-                measurements (dict): Measured qubits/classical bits map and number of classical bits
-                number_of_qubits (int): number of qubits used in the algorithm
-                number_of_shots (int): The number of times the algorithm is executed.
+        Args:
+            result (dict): The result output from the quantum inspire backend with full-
+                           state projection histogram output.
+            measurements (dict): list of measured qubits/classical bits map and number of classical bits
```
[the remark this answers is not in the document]
qSaevar: This holds true in other docstrings as well.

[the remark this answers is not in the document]
QFer: I'll go through all the docstrings of this file and make it consistent.

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -171,8 +178,44 @@ def test_get_experiment_results_returns_correct_value(self):
...
+    def test_get_experiment_results_returns_single_shot(self):
+        number_of_shots = 1
+        self._basic_job_dictionary['number_of_shots'] = number_of_shots
+        instructions = [{'name': 'h', 'params': [], 'texparams': [], 'qubits': [0]},
+                        {'name': 'cx', 'params': [], 'texparams': [], 'qubits': [0, 1]},
+                        {'name': 'measure', 'qubits': [1], 'memory': [1]},
+                        {'name': 'measure', 'qubits': [0], 'memory': [0]}]
+        experiment = self._instructions_to_two_qubit_experiment(instructions)
+        api = Mock()
+        api.get.side_effect = self.__api_get_return_values_single_shot
```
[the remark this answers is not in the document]
QFer: with return_value set, each get on the api gets the same return value, with side_effect the function is called and there the url can be checked to see if the right url was passed, or depending on the url another return value can be returned when get was called more than once within the test. For the last reason I introduced this because get was called for getting the result and for getting the raw_data. Since the latest update, the get methode is now called once because we made an get_raw_data methode on the api to get the raw_data, so setting return_value is enough.

QFer:
qSaevar, I rebased to dev and had to do a git push --force

PR #28, opened 2019-02-20
Generating count histogram from the raw data

QFer:
With this issue, the raw data that is returned from the backend is taken as the source to calculate the count histogram (instead of the probabilities histogram). Also the memory data field is filled with the raw data values.
A special case is when number_of_shots = 1. In this case the backend doesn't return raw_data.
With number_of_shots = 1, there must be one value in the counts histogram (with count = 1), and the same single value in memory data. To determine this value a random float is generated in the range [0, 1). With this random number the value from the probabilities histogram is taken where the added probabilities is greater this random number.

Also the unit tests are adjusted because the single shot unit test now delivers valid data.

Some minor docstring adjustments for the other part of the file.

### DIFF
--- a/src/quantuminspire/qiskit/backend_qx.py
+++ b/src/quantuminspire/qiskit/backend_qx.py
@@ -18,7 +18,8 @@ limitations under the License.
 import io
 import json
 import uuid
-from collections import defaultdict, OrderedDict
+import numpy as np
+from collections import defaultdict, OrderedDict, Counter
 
 from coreapi.exceptions import ErrorMessage
 from qiskit.providers import BaseBackend
@@ -106,8 +107,7 @@ class QuantumInspireBackend(BaseBackend):
         return job
 
     def retrieve_job(self, job_id):
-        """
-        Retrieve a specified job by its job_id.
+        """ Retrieve a specified job by its job_id.
 
         Args:
             job_id (str): The job id.
@@ -160,8 +160,8 @@ class QuantumInspireBackend(BaseBackend):
         return job_id
 
     def get_experiment_results(self, qi_job):
-        """
-        Get results from experiments from the Quantum-inspire platform.
+        """ Get results from experiments from the Quantum-inspire platform.
+
         Args:
             qi_job (QIJob): A job that has already been submitted and which execution is completed.
 
@@ -169,7 +169,7 @@ class QuantumInspireBackend(BaseBackend):
             QisKitBackendError: If an error occurred during execution by the backend.
 
         Returns:
-            List: A list of experiment results; containing the data, execution time, status, etc.
+            list: A list of experiment results; containing the data, execution time, status, etc.
         """
         jobs = self.__api.get_jobs_from_project(qi_job.job_id())
         results = [self.__api.get(job['results']) for job in jobs]
@@ -181,20 +181,11 @@ class QuantumInspireBackend(BaseBackend):
 
             user_data = json.loads(job.get('user_data'))
             measurements = user_data.pop('measurements')
-            histogram, full_state_histogram = QuantumInspireBackend.__convert_histograms(result, measurements,
-                                                                                         result['number_of_qubits'],
-                                                                                         job['number_of_shots'])
-            histogram_obj = Obj.from_dict(histogram)
-            full_state_histogram_obj = Obj.from_dict(full_state_histogram)
-            raw_data = self.__api.get_raw_data(str(result['id']))
-            memory_data = self.__get_memory_data(raw_data, measurements, result['number_of_qubits'])
-            if memory_data:
-                experiment_result_data = ExperimentResultData(counts=histogram_obj,
-                                                              probabilities=full_state_histogram_obj,
-                                                              memory=memory_data)
-            else:
-                experiment_result_data = ExperimentResultData(counts=histogram_obj,
-                                                              probabilities=full_state_histogram_obj)
+            histogram_obj, memory_data = self.__convert_result_data(result, measurements)
+            full_state_histogram_obj = self.__convert_histogram(result, measurements)
+            experiment_result_data = ExperimentResultData(counts=histogram_obj,
+                                                          probabilities=full_state_histogram_obj,
+                                                          memory=memory_data)
             header = Obj.from_dict(user_data)
             experiment_result_dictionary = {'name': job.get('name'), 'seed': 42, 'shots': job.get('number_of_shots'),
                                             'data': experiment_result_data, 'status': 'DONE', 'success': True,
@@ -270,7 +261,7 @@ class QuantumInspireBackend(BaseBackend):
             experiment (QobjExperiment): The experiment with gate operations and header.
 
         Returns:
-            List: A list of lists, for each measurement the returned list contains a list of
+            list: A list of lists, for each measurement the returned list contains a list of
                   [qubit_index, classical_bit_index], which represents the measurement of a qubit to a classical bit.
         """
         header = experiment.header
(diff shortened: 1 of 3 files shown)


src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -308,58 +299,84 @@ def __qubit_to_classical_hex(qubit_register, measurements, number_of_qubits):
         return classical_state_hex
 
     @staticmethod
-    def __get_memory_data(raw_data, measurements, number_of_qubits):
-        """ The quantum inspire backend returns the single shot values. This function
-            converts the raw data to hexadecimal memory data according the Qiskit spec.
-
-        Note:
-            When shots = 1, the backend returns an empty list as raw_data.
+    def __convert_histogram(result, measurements):
```
qSaevar: It all about typehint these days

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -308,58 +299,84 @@ def __qubit_to_classical_hex(qubit_register, measurements, number_of_qubits):
...
-            output_histogram_counts[classical_state_hex] += int(probability * number_of_shots)
             output_histogram_probabilities[classical_state_hex] += probability
 
-        return (OrderedDict(sorted(output_histogram_counts.items(), key=lambda kv: int(kv[0], 16))),
-                OrderedDict(sorted(output_histogram_probabilities.items(), key=lambda kv: int(kv[0], 16))))
+        full_state_histogram_obj = OrderedDict(sorted(output_histogram_probabilities.items(),
+                                                      key=lambda kv: int(kv[0], 16)))
+        return Obj.from_dict(full_state_histogram_obj)
+
+    def __convert_result_data(self, result, measurements):
```
qSaevar: typehints

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -308,58 +299,84 @@ def __qubit_to_classical_hex(qubit_register, measurements, number_of_qubits):
...
+        memory_data = []
+        histogram_data = defaultdict(lambda: 0)
+        number_of_qubits = result['number_of_qubits']
+        raw_data = self.__api.get_raw_data(str(result['id']))
+        if raw_data:
+            for qubit_register in raw_data:
+                classical_state_hex = QuantumInspireBackend.__qubit_to_classical_hex(qubit_register, measurements,
+                                                                                     number_of_qubits)
+                memory_data.append(classical_state_hex)
+            for elem, count in Counter(memory_data).items():
```
qSaevar: this could be a dict comprehension right?

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -308,58 +299,84 @@ def __qubit_to_classical_hex(qubit_register, measurements, number_of_qubits):
...
+                histogram_data[elem] = count
+        else:
+            state_probabilities = result['histogram']
+            prob = np.random.rand()
```
qSaevar: `random_probability` maybe?

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -206,11 +206,17 @@ def test_get_experiment_results_returns_single_shot(self):
         job = QIJob('backend', '42', api)
         simulator = QuantumInspireBackend(api, Mock())
         experiment_result = simulator.get_experiment_results(job)[0]
```
qSaevar: would be nice to have test that runs this few thousand times, with a bit more complex histogram, to see if the probability is similar.

Say, if the histogram is ` {'0': 0.2, '1': 0.3, '2': 0.4, '3': 0.1}` and we run this 10000 times, then the output of that test should be something like: ` {'0x2': 4015, '0x1': 2989, '0x3': 985, '0x0': 2011})`

src/quantuminspire/qiskit/backend_qx.py
QFer: ok

src/quantuminspire/qiskit/backend_qx.py
QFer: yep

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Added test_get_experiment_results_multiple_single_shots which tests your suggestion

src/quantuminspire/qiskit/backend_qx.py
QFer: I'll add them

src/quantuminspire/qiskit/backend_qx.py
QFer: I'll add them for the file

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -203,11 +200,11 @@ def get_experiment_results(self, qi_job):
         return experiment_results
 
     @staticmethod
-    def __validate(job):
+    def __validate(job: Qobj) -> None:
         """ Validates the number of shots, classical bits and compiled Qiskit circuits.
 
         Args:
-            job (QObj): The quantum job with the Qiskit algorithm and quantum inspire backend.
+            job (Qobj): The quantum job with the Qiskit algorithm and quantum inspire backend.
```
qSaevar: Now that typehints are in the signature, they should not also be in the docstring.

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -206,14 +206,55 @@ def test_get_experiment_results_returns_single_shot(self):
...
+            job = QIJob('backend', '42', api)
+            simulator = QuantumInspireBackend(api, Mock())
+            experiment_result = simulator.get_experiment_results(job)[0]
+            # Exactly one value in memory
+            self.assertEqual(len(experiment_result.data.memory), 1)
+            # The only value in memory is the same as the value in the counts histogram.
+            self.assertEqual(list(experiment_result.data.counts.to_dict().keys())[0], experiment_result.data.memory[0])
+            one_shot_results[experiment_result.data.memory[0]] += 1
+
+        self.assertIn(one_shot_results['0x0'], range(1850, 2150))
```
qSaevar: if you put a `random.seed(seed)` in the test, then you always get the exact same number back. Can be handy.

src/quantuminspire/qiskit/backend_qx.py
QFer: Done. Ran in some trouble with recursive imports (QuantumInspireProver -> QuantumInspireBackend -> QuantumInspireProvider) for the type hint in __init__ from QuantumInspireBackend. Solved it in Quantum_Inspire_Provider.py so there's one extra file.

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Handy indeed. The result of unittest was now not 100% ok. Adjusted the unittest.

src/quantuminspire/qiskit/quantum_inspire_provider.py
```diff
@@ -32,6 +31,8 @@ def backends(self, name=None, **kwargs):
         Returns:
             list<QuantumInspireBackend>: List of backends that meet the filter requirements.
         """
+        from quantuminspire.qiskit.backend_qx import QuantumInspireBackend
```
qSaevar: PEP8 is not happy about this, so, I think we need to relax a bit on the typehint in the backend. By, e.g.,  using BaseProvider or any even though it is not truly true.

QFer: Used 'Any' to get rid of recursive imports. Docstring now point out the right type again.

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -262,16 +259,17 @@ def __validate_no_gates_after_measure(experiment):
                     raise QisKitBackendError('Operation after measurement!')
 
     @staticmethod
-    def _collect_measurements(experiment):
+    def _collect_measurements(experiment: QobjExperiment) -> Dict:
```
qSaevar: Should we hint at the types in dict?

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -286,18 +284,17 @@ def _collect_measurements(experiment):
         return {'measurements': measurements, 'number_of_clbits': number_of_clbits}
 
     @staticmethod
-    def __qubit_to_classical_hex(qubit_register, measurements, number_of_qubits):
+    def __qubit_to_classical_hex(qubit_register: int, measurements: Dict, number_of_qubits: int) -> str:
```
qSaevar: Do we know more about measurements type, other than it's just a dict?

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -308,58 +305,82 @@ def __qubit_to_classical_hex(qubit_register, measurements, number_of_qubits):
         return classical_state_hex
 
     @staticmethod
-    def __get_memory_data(raw_data, measurements, number_of_qubits):
-        """ The quantum inspire backend returns the single shot values. This function
-            converts the raw data to hexadecimal memory data according the Qiskit spec.
-
-        Note:
-            When shots = 1, the backend returns an empty list as raw_data.
+    def __convert_histogram(result: Dict, measurements: Dict) -> Obj:
```
qSaevar: Guess the comments above, or the questions, apply to all container hints.

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -49,13 +53,14 @@ class QuantumInspireBackend(BaseBackend):
         max_shots=1024
     )
 
-    def __init__(self, api, provider, configuration=None):
+    def __init__(self, api: QuantumInspireAPI, provider: Any,
```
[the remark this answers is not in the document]
QFer: DEM-674 extended

src/quantuminspire/qiskit/backend_qx.py
QFer: Added type hint for Dict

src/quantuminspire/qiskit/backend_qx.py
QFer: Added type hint for Dict

src/quantuminspire/qiskit/backend_qx.py
QFer: Added type hint for Dict

QFer:
The OrderedDict of project parameter I'll have to do another time when type hinting it together with Api

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -308,58 +306,84 @@ def __qubit_to_classical_hex(qubit_register, measurements, number_of_qubits):
...
 
     @staticmethod
-    def __get_memory_data(raw_data, measurements, number_of_qubits):
-        """ The quantum inspire backend returns the single shot values. This function
-            converts the raw data to hexadecimal memory data according the Qiskit spec.
-
-        Note:
-            When shots = 1, the backend returns an empty list as raw_data.
+    def __convert_histogram(result: Dict[str, Union[int, str, float, Dict[str, float]]],
+                            measurements: Dict[str, Union[List[List[int]], int]]) -> Obj:
```
qSaevar: guess I asked for this...:/

src/quantuminspire/qiskit/quantum_inspire_provider.py
```diff
@@ -1,36 +1,37 @@
...
-from quantuminspire.qiskit.backend_qx import QuantumInspireBackend
 
 QI_URL = 'https://api.quantum-inspire.com'
 
 
 class QuantumInspireProvider(BaseProvider):
     """ Provides a backend and an api for a single Quantum Inspire account. """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
```
qSaevar: type of `*args` and `**kwargs`?

src/quantuminspire/qiskit/quantum_inspire_provider.py
```diff
@@ -1,36 +1,37 @@
...
         super().__init__(*args, **kwargs)
         self._backends = []
         self._api = None
 
-    def __str__(self):
+    def __str__(self) -> str:
         return 'QI'
 
-    def backends(self, name=None, **kwargs):
+    def backends(self, name: str = None, **kwargs) -> List[QuantumInspireBackend]:
```
qSaevar: again, should we not also typehint the `**kwargs`?

src/quantuminspire/qiskit/backend_qx.py
QFer: It's now maybe more than a hint...

src/quantuminspire/qiskit/quantum_inspire_provider.py
QFer: Implementation of super is:
    def __init__(self, *args, **kwargs):
        pass
so, it can be Any....

src/quantuminspire/qiskit/quantum_inspire_provider.py
QFer: comments says: **kwargs: Used for filtering, not implemented.
So it can be Any-thing...

PR #31, opened 2019-03-01
[DEM-NNN] type hinting

QFer:
Type hints added in all files but projectq backend because it is in progress in another branch.
Mypy is satisfied except for missing external imports (Class cannot subclass 'blabla' (has type 'Any'))

### DIFF
--- a/src/quantuminspire/job.py
+++ b/src/quantuminspire/job.py
@@ -10,7 +10,7 @@ class QuantumInspireJob:
             methods to check the status and retrieve the results from the API.
 
         Arguments:
-            api (QuantumInspireApi): An instance to the API.
+            api: An instance to the API.
             job_identifier: The job identification number.
         """
         QuantumInspireJob.__check_arguments(api, job_identifier)
--- a/src/quantuminspire/job.py
+++ b/src/quantuminspire/job.py
@@ -1,27 +1,29 @@
+from typing import Dict, Any
+from collections import OrderedDict
 from coreapi.exceptions import ErrorMessage
 
 
 class QuantumInspireJob:
 
-    def __init__(self, api, job_identifier):
-        """ The QuantumInstpire Job class encapsulates the base job of the API and has
-            methods to check the status and retieve the results from the API.
+    def __init__(self, api: Any, job_identifier: int) -> None:
+        """ The QuantumInspire Job class encapsulates the base job of the API and has
+            methods to check the status and retrieve the results from the API.
 
         Arguments:
             api (QuantumInspireApi): An instance to the API.
-            job_identifier (int): The job identification number.
+            job_identifier: The job identification number.
         """
         QuantumInspireJob.__check_arguments(api, job_identifier)
-        self.__job_identifier = job_identifier
-        self.__api = api
+        self.__job_identifier: int = job_identifier
+        self.__api: Any = api
 
     @staticmethod
-    def __check_arguments(api, job_identifier):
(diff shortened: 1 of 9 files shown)


src/quantuminspire/job.py
```diff
@@ -1,27 +1,29 @@
...
 
-    def __init__(self, api, job_identifier):
-        """ The QuantumInstpire Job class encapsulates the base job of the API and has
-            methods to check the status and retieve the results from the API.
+    def __init__(self, api: Any, job_identifier: int) -> None:
+        """ The QuantumInspire Job class encapsulates the base job of the API and has
+            methods to check the status and retrieve the results from the API.
 
         Arguments:
             api (QuantumInspireApi): An instance to the API.
```
qSaevar: Here is still a docstring typehint. And was there a problem with the new typehint for the api?

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -30,12 +30,12 @@
...
 from quantuminspire.qiskit.qi_job import QIJob
-from quantuminspire.version import __version__ as quantum_inspire_version
-from quantuminspire.job import QuantumInspireJob
 from quantuminspire.api import QuantumInspireAPI
+from quantuminspire.exceptions import QisKitBackendError
+from quantuminspire.job import QuantumInspireJob
+from quantuminspire.version import __version__ as quantum_inspire_version
 
 
 class QuantumInspireBackend(BaseBackend):
```
qSaevar: There are still a lot of typehints int the __init__ docstring

src/quantuminspire/qiskit/qi_job.py
```diff
@@ -16,46 +18,46 @@ class QIJob(BaseJob):
             result = job.result()
     """
 
-    def __init__(self, backend, job_id, api, qobj=None):
+    def __init__(self, backend: Any, job_id: str, api: QuantumInspireAPI, qobj: Optional[Qobj] = None) -> None:
         """
         Construct a new QIJob object. Not normally called directly, use a backend object to create/retrieve jobs.
 
         Args:
             backend (QuantumInspireBackend): A quantum-inspire backend.
```
qSaevar: typehint in docstring

QFer:
Because of the flag --ignore-missing-imports any imported module that cannot be found is silently replaced with Any. That's causing this mypy unhappiness. I solved it by silencing the checker on these spots (https://mypy.readthedocs.io/en/latest/common_issues.html)

I left the docstring types for those types that caused recursive import problems [DEM-674], because Any is not very informative, but I deleted this now. The docstring types in __init__ is the description of the used configuration. As the docstring descriptions in api.py, this is not replaced by type hinting because these structures/classes are not ours.

src/quantuminspire/qiskit/backend_qx.py
QFer: The docstring types in init is the description of the used configuration. As the docstring descriptions in api.py, this is not replaced by type hinting because these structures/classes are not ours.

src/quantuminspire/qiskit/qi_job.py
QFer: I left the docstring types for those types that caused recursive import problems [DEM-674], because Any is not very informative, but I deleted this now.

src/quantuminspire/job.py
QFer: I left the docstring types for those types that caused recursive import problems [DEM-674], because Any is not very informative, but I deleted this now.

PR #32, opened 2019-03-06
[DEM-722] Jenkins build

QFer:
First commit for enabling:
* To run SDK against qiskit's latest development version (called master) from this PR/branch

When this PR is merged, Jenkins has to be adjusted (configured) to build the master branch of quantuminspire instead of this PR.


QFer:
The Jenkinsfile includes:
1. installing the latest qiskit from mypy (installing qiskit, qiskit-aer and qiskit-terra).
2. uninstalling the qiskit package (leaving terra and aer)
3. cloning qiskit from github master and installing it (the develop version)
4. Installing quantuminspire from the repository
5. Setting the credentials (API_USERNAME, API_PASSWORD)
6. Running the qiskit example file which uses these credentials

I currently left out the unittesting (and mypy-ing) of quantuminspire to detect errors in these tests because it's not in the goal of this ticket. Of course we can add them.

docs/example_qiskit_entangle.py
```diff
@@ -21,13 +22,19 @@
...
+QI_PASSWORD = os.getenv('API_PASSWORD')
+
 
 def get_authentication():
     """ Gets the authentication for connecting to the Quantum Inspire API."""
-    print('Enter email:')
-    email = input()
-    print('Enter password')
-    password = getpass()
+    if QI_EMAIL and QI_PASSWORD:
```
quantumkoen: This is a bit implicit, we should be more explicit here I think
```suggestion
    if QI_EMAIL is not None and QI_PASSWORD is not None:
```

src/tests/Jenkinsfile
```diff
@@ -1,33 +1,64 @@
...
-            steps {
                 sh '''
-                       . ./venv/bin/activate &&
-                       coverage run --source="./src/quantuminspire" -m unittest discover -s src/tests -t src -v &&
-                       coverage report -m &&
-                       coverage xml
+                    . ./venv/bin/activate &&
+                    pwd &&
+                    ls -lah &&
+                    pip install qiskit &&
```
quantumkoen: Why install qiskit and then uninstall it again?

If there is a valid reason for doing this, it would be good to add a comment I think, because it is unexpected.

Regardless, should probably include `--cache-dir /var/pip_cache` and possibly a `--upgrade` here to avoid all the `The directory '/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.` errors?

src/tests/Jenkinsfile
```diff
@@ -1,33 +1,64 @@
...
+                    git clone https://github.com/Qiskit/qiskit &&
+                    pip install ./qiskit -r ./qiskit/requirements-dev.txt
                    '''
             }
         }
-        stage('Coverage') {
+        stage('Build') {
+            steps {
+                sh 'virtualenv venv'
+                sh '. ./venv/bin/activate && pip install . --cache-dir /var/pip_cache -r ./src/tests/requirements.txt'
```
quantumkoen: We should add a `--upgrade` to the `pip install` probably, to make sure we use the latest dependencies. If we don't the cached versions will be used, which may be old versions.

src/tests/Jenkinsfile
```diff
@@ -1,33 +1,64 @@
...
+                        sh 'echo "API_URL: ${API_URL}"'
+                        sh 'echo "API_CREDENTIALS_ID: ${API_CREDENTIALS_ID}"'
+                        sh 'echo "API_USERNAME: ${API_USERNAME}"'
+                        sh 'echo "API_PASSWORD: ${API_PASSWORD}"'
+                        sh '. ./venv/bin/activate && python ./docs/example_qiskit_entangle.py'
+                    }
+                }
             }
         }
+        //stage('Collect') {
```
quantumkoen: Commented out code should be removed.

src/tests/Jenkinsfile
```diff
@@ -1,33 +1,64 @@
...
-                       coverage run --source="./src/quantuminspire" -m unittest discover -s src/tests -t src -v &&
-                       coverage report -m &&
-                       coverage xml
+                    . ./venv/bin/activate &&
+                    pwd &&
+                    ls -lah &&
+                    pip install qiskit &&
+                    pip uninstall -y qiskit &&
+                    git clone https://github.com/Qiskit/qiskit &&
+                    pip install ./qiskit -r ./qiskit/requirements-dev.txt
```
quantumkoen: Should probably include `--cache-dir /var/pip_cache` and possibly a `--upgrade` here?

src/tests/Jenkinsfile
```diff
@@ -1,33 +1,64 @@
...
+        stage('Build') {
+            steps {
+                sh 'virtualenv venv'
+                sh '. ./venv/bin/activate && pip install . --cache-dir /var/pip_cache -r ./src/tests/requirements.txt'
+           }
+        }
+        stage('Backend Test') {
             steps {
-                step([$class: 'CoberturaPublisher', coberturaReportFile: 'coverage.xml'])
+                script {
```
quantumkoen: You don't actually need to persist the workspace here, since you are not accessing anything in a subsequent step. It doesn't harm, but if it is not needed better to just remove it to avoid confusion.

docs/example_qiskit_entangle.py
QFer: oke

src/tests/Jenkinsfile
QFer: oke, i'll add cache dir and upgrade

src/tests/Jenkinsfile
QFer: I'll remove it

src/tests/Jenkinsfile
QFer: Added comment:
# Installing qiskit, qiskit-aer and qiskit-terra as separated (development) packages doesn't work as expected.
# As a workaround we install the qiskit packages from PyPI, uninstall package qiskit and reinstall
# the development version of qiskit afterwards.

src/tests/Jenkinsfile
QFer: oke.

QFer:
In the ticket [DEM-722] are some todo's after the review comments are solved.

src/tests/Jenkinsfile
```diff
@@ -1,32 +1,59 @@
...
+                    . ./venv/bin/activate &&
+                    pwd &&
+                    ls -lah &&
+                    # Installing qiskit, qiskit-aer and qiskit-terra as separated (development) packages doesn't work
+                    # as expected. As a workaround we install the qiskit packages from PyPI, uninstall package qiskit
+                    # and reinstall the development version of qiskit afterwards.
+                    pip install qiskit --cache-dir /var/pip_cache &&
+                    pip uninstall -y qiskit --cache-dir /var/pip_cache &&
+                    git clone https://github.com/Qiskit/qiskit &&
+                    pip install ./qiskit --upgrade --cache-dir /var/pip_cache -r ./qiskit/requirements-dev.txt
```
quantumkoen: This only installs the meta package, you can probably leave this out, but add `pip uninstall qiskit-terra` and then clone qiskit terra and install manually. Maybe the same for qiskit-aer and qiskit-ignis, although I think they are less important for the goal of this test (but then again, you might get incompatibility issues if you only install master of terra, and not master of the rest).

QFer: Same on the build server with Jenkins while installing qiskit-aer: Could NOT find BLAS (missing: BLAS_LIBRARIES). I'll try without uninstalling qiskit-aer.

quantumkoen: Yeah you have to `apt install libblas-dev`, it's not installed by default. But not installing qiskit-aer is fine, since we don't use it for our tests anyway.

QFer: I tried but it needs administrator rights to install which I don't have. I'll leave it out for now.

PR #33, opened 2019-03-07
[DEM-811] api._get made private

QFer:
* Api.get was 'misused' for getting information from the backed. Now it is used once for getting the schema only.
* Api._get is now private method of Api
* Added public methods to Api to get the available information from the backend
* Restructured the unit tests for the code and functions where api.get was used and now another api-method is used.
* Added unit tests for the Api methods I added to get data from the backend now the get method is private.

### DIFF
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -61,7 +61,7 @@ class QuantumInspireAPI:
                                   Default set to coreapi.Client.
 
         Note: When no project name is given, a temporary project is created for the job and deleted after the job
-              has finished. When a project name is given, a project is created if it does not exists, but re-used
+              has finished. When a project name is given, a project is created if it does not exist, but re-used
               if a project with that name already exists. In either case, the project will not be deleted when a
               project name is supplied here.
 
@@ -76,7 +76,7 @@ class QuantumInspireAPI:
         except (CoreAPIException, TypeError) as ex:
             raise ApiError('Could not connect to {}'.format(base_uri)) from ex
 
-    def get(self, uri_path: str) -> Any:
+    def _get(self, uri_path: str) -> Any:
         """ Method for making requests to the coreapi client instance to get some piece of information. The information
             requested depends on the uri_path parameter.
 
@@ -117,7 +117,7 @@ class QuantumInspireAPI:
 
     def _load_schema(self) -> None:
         """ Loads the schema with metadata that explains how the api-data is structured."""
-        self.document = self.get(urljoin(self.base_uri, 'schema/'))
+        self.document = self._get(urljoin(self.base_uri, 'schema/'))
 
     def list_backend_types(self) -> None:
         """ Prints the backend types with the name and the maximum number of qubits it supports."""
@@ -181,7 +181,7 @@ class QuantumInspireAPI:
             backend_name: The backend name.
 
         Raises:
-            ApiError: An ApiError exception is thrown when the backend name does not exists.
+            ApiError: An ApiError exception is thrown when the backend name does not exist.
 
         Returns:
             The properties of the backend type of the specific backend.
@@ -202,7 +202,7 @@ class QuantumInspireAPI:
             identifier: The backend type identifier.
 
         Raises:
-            ApiError: If the requested backend type does not exists.
+            ApiError: If the requested backend type does not exist.
             ValueError: If the backend type identifier is not of the correct type.
 
         Returns:
@@ -235,7 +235,7 @@ class QuantumInspireAPI:
             project_id: The project identification number.
 
         Raises:
-            ApiError: If the requested project does not exists.
+            ApiError: If the requested project does not exist.
 
         Returns:
             The properties describing the project:
@@ -293,7 +293,7 @@ class QuantumInspireAPI:
             project_id: The project identification number.
 
         Raises:
-            ApiError: If the project identified by project_id does not exists.
+            ApiError: If the project identified by project_id does not exist.
         """
         payload = {
             'id': project_id
@@ -319,7 +319,7 @@ class QuantumInspireAPI:
             job_id: The job identification number.
 
         Raises:
-            ApiError: If the requested job does not exists.
+            ApiError: If the requested job does not exist.
 
         Returns:
             The properties describing the job:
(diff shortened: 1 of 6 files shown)


src/quantuminspire/api.py
```diff
@@ -546,6 +646,51 @@ def get_assets(self) -> List[Dict[str, Any]]:
...
+
+        Returns:
+            The assets with all of its properties.
+            See `get_asset` for a description of the asset properties.
+        """
+        job = self.get_job(job_id)
+        asset_url = str(job.get('input'))
+        try:
+            asset_id = int(asset_url.split('/')[-2])
+        except (ValueError, IndexError) as err_msg:
```
qSaevar: In above methods, only IndexError is excepted, not ValueError. Is this a special case or should ValueError also be included above? And, since we are on it, should this maybe be extracted since it is the same-ish block over and over again?

src/quantuminspire/api.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, base_uri: str, authentication: coreapi.auth.BasicAuthenticati
         except (CoreAPIException, TypeError) as ex:
             raise ApiError('Could not connect to {}'.format(base_uri)) from ex
 
-    def get(self, uri_path: str) -> Any:
+    def _get(self, uri_path: str) -> Any:
```
qSaevar: get should be private according to the ticket, i.e., __get not _get.

src/quantuminspire/api.py
```diff
@@ -482,24 +502,104 @@ def get_results(self) -> List[Dict[str, Any]]:
...
+    def get_raw_data_from_result(self, result_id: int) -> List[int]:
         """ Gets the raw data from the result of the executed cQASM code, given the result_id. The raw data consists
             of a list with integer state values for each shot of the experiment (see job.number_of_shots).
 
         Args:
             result_id: The identification number of the result.
 
         Raises:
-            ApiError: If the requested result with identification result_id does not exists.
+            ApiError: If the requested result with identification result_id does not exist.
```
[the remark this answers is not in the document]
QFer: Adjusted to:
ApiError: If the raw data url in result is invalid or the request for the raw data using the url failed.

src/quantuminspire/api.py
```diff
@@ -546,6 +646,51 @@ def get_assets(self) -> List[Dict[str, Any]]:
...
+            An empty list is returned when the project has no assets.
+
+        Raises:
+            ApiError: If the project identified by project_id does not exist.
+        """
+        try:
+            assets = self._action(['projects', 'assets', 'list'], params={'id': project_id})
+        except ErrorMessage as err_msg:
+            raise ApiError('Project with id {} does not exist!'.format(project_id)) from err_msg
+        ret: List[Dict[str, Any]] = assets
```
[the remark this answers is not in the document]
QFer: Ok

src/quantuminspire/api.py
QFer: The ValueError is added because in this case the asset_id is an integer and has to be converted from a substring fromthe url. In the other cases the token was a string.

I thought about extracting a piece of code but I don't think we win something because except the exception framework the variables and error messages are different so we will have to put them all in parameters.

src/quantuminspire/api.py
QFer: One underscore is considered as a private method/instance variable in python. With two underscores the name is mangled. I don't see why that is necessary. Further, all the other private methods within this class have one underscore (_action, _load_schema, _create_asset etc.)

src/quantuminspire/api.py
```diff
@@ -358,6 +358,26 @@ def get_jobs(self) -> List[Dict[str, Any]]:
         ret: List[Dict[str, Any]] = self._action(['jobs', 'list'])
         return ret
 
+    def get_jobs_from_asset(self, asset_id: int) -> List[Dict[str, Any]]:
+        """ Gets the jobs with its properties for a asset, given the asset id.
```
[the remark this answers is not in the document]
QFer: ok

QFer:
Rebased to dev, had to force push.

src/quantuminspire/api.py
qSaevar: I agree, I think single underscore is enough since that means, in python world, that it should not be accessed outside the class. But in our guidelines it states that __ is private and _ is protected.
But...
To clarify I talked to Koen and he ment it should "not be public" so I'm happy with the single underscore.

src/tests/quantuminspire/test_api.py
```diff
@@ -513,20 +640,116 @@ def test_get_result_RaisesApiError(self):
...
+        self.assertRaisesRegex(ApiError, 'Invalid raw data url for result with', api.get_raw_data_from_result,
+                               result_id=result_identity)
+
+    def test_get_raw_data_invalid_from_result_RaisesApiError(self):
+        result_identity = 486
+        expected_payload = {'id': result_identity, 'token': '162c'}
+        self.coreapi_client.handlers['results'] = partial(self.__mock_errors_in_result_handler,
+                                                          expected_payload, 'read')
+        api = QuantumInspireAPI('FakeURL', self.authentication, coreapi_client_class=self.coreapi_client)
+        self.assertRaisesRegex(ApiError, 'Raw data for result with id', api.get_raw_data_from_result,
```
qSaevar: Why leave the rest of the error message out? ...486 does not exist!

QFer: Whole message added

src/quantuminspire/job.py
```diff
@@ -53,9 +52,8 @@ def retrieve_results(self) -> Dict[str, Any]:
             histogram of the job. When an error has occurred the raw_text item shall not be
             an empty string.
         """
-        job = self.__api.get_job(self.__job_identifier)
-        result_uri = job['results']
-        return OrderedDict(self.__api.get(result_uri))
+        result: Dict[str, Any] = self.__api.get_result_from_job(self.__job_identifier)
```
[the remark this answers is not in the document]
QFer: The statement:
return self.__api.get_result_from_job(self.__job_identifier)
will result in a mypy warning:
quantuminspire\job.py:57: warning: Returning Any from function declared to return "Dict[str, Any]"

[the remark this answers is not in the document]
qSaevar: Ahhh, because the api is any

PR #35, opened 2019-03-14
[DEM-743] examples installed

QFer:
After discussion we decided to install the examples in the directory: share/doc/quantuminspire/examples in the library root, depending on `sys.prefix` or `site.USER_BASE`. The README.md will hint to this directory with the mentioning of `sys.prefix` or `site.USER_BASE`.


QFer:
the example files are added individually because when the docs directory is used for other stuff than examples we don't put this all in the examples directory. On the other hand, when an example is added now we don't have to forget to install it and add it to setup.py.

quantumkoen:
> the example files are added individually because when the docs directory is used for other stuff than examples we don't put this all in the examples directory. On the other hand, when an example is added now we don't have to forget to install it and add it to setup.py.

Of course, we could move the examples to `docs/examples/` and then glob that.....

PR #37, opened 2019-03-19
[DEM-723] Added the long description from the README in the setup file.

ammar92:


setup.py
```diff
@@ -31,8 +30,19 @@ def get_version_number(module):
     return version_number
 
 
+def get_long_description():
+    """ Extract the long description from the README file """
+
+    with open('README.md') as f:
```
quantumkoen:
```suggestion
    with open('README.md', encoding='utf-8') as f:
```

PR #40, opened 2019-03-26
[DEM-806] Binary controlled gates

QFer:
* Parameter 'conditional' on the backend made True
* Generating the cqasm now via public method parse instead of calling the private methods from the circuit_parser
* Circuit parser parse method - checks if the instruction is a conditional or a 'normal'.
* In qiskit a conditional gate operation is based on an integer value of the classical registers, in cQASM the conditional gate is check against the positive value of bits. c_if(c, 11) actually means bit 2 and 4 must be false and bits 0,1 and 3 must be true. Because cQASM only checks for positive values of bits, bit 2 and 4 must be inverted before and after the gate operation.
* For each supported gate, a conditional gate cQASM generator is added.
* Unittests added for the new methods. All the possible exceptions are tested.
* Added the identity operator generating methods because it was in the basis_gates but not implemented.


src/quantuminspire/qiskit/circuit_parser.py
```diff
@@ -191,60 +431,135 @@ def _u3(self, circuit: Dict[str, Any]) -> str:
...
+        masked_val = mask & val
+
+        # form the negation to the 0-values of the measurement registers, when mask == val no bits are negated
+        negate_zeroes_line = ''
+        if masked_val != mask:
+            negate_zeroes_line = 'not b[' + ','.join(
+                str(i) for i in range(nr_of_classical_bits) if not (masked_val & (1 << i))) + ']\n'
+
+        # form multi bits control - qasm-single-gate-multiple-qubits
+        binary_control = 'b[0:{}], '.format(nr_of_classical_bits - 1)
```
qSaevar: does binary control always start at index '0'? is this not allowed `C-H b[2:3], q[1]`? And, if `nr_of_classical_bits` is 1, then this will become `b[0:0]` which is not very pretty and can not be executed from the web interface.

src/tests/quantuminspire/qiskit/test_circuit_parser.py
```diff
@@ -130,24 +184,45 @@ def test_generate_cqasm_CorrectOutputGateU1(self):
...
+        result = self._generate_cqasm_from_instructions(instructions, 2)
+        self.assertFalse('q[0]' in result)
+
+    def test_generate_cqasm_correct_output_gate_u2(self):
         instructions = [{'name': 'u2', 'qubits': [0], 'params': [np.pi, np.pi / 2],
                          'texparams': ['\\pi', '\\frac{\\pi}{2}']}]
         result = self._generate_cqasm_from_instructions(instructions, 2)
         self.assertTrue('Rz q[0], 1.570796\nRy q[0], 1.570796\nRz q[0], 3.141593\n' in result)
 
+        instructions[0]['conditional'] = {'mask': '0xF', 'type': 'equals', 'val': '0x3'}
```
qSaevar: Should the conditional gates not have their own test methods?

src/tests/quantuminspire/qiskit/test_circuit_parser.py
```diff
@@ -197,11 +297,48 @@ def test_generate_cqasm_CorrectOutputRotationY(self):
...
         result = self._generate_cqasm_from_instructions(instructions, 2)
         self.assertTrue('Rz q[1], -1.570796\n' in result)
+
+        instructions[0]['conditional'] = {'mask': '0xF', 'type': 'equals', 'val': '0x1'}
+        result = self._generate_cqasm_from_instructions(instructions, 2)
+        self.assertTrue('not b[1,2,3]\nC-Rz b[0:3], q[1], -1.570796\nnot b[1,2,3]\n' in result)
+
+    def test_generate_cqasm_correct_output_unknown_gate(self):
+        instructions = [{'name': 'bla', 'qubits': [1], 'params': [-np.pi / 2]}]
+        self.assertRaisesRegex(ApiError, 'Gate bla not supported', self._generate_cqasm_from_instructions,
```
qSaevar: It is about time that we start supporting the bla, and c-bla, gates.

src/quantuminspire/qiskit/circuit_parser.py
```diff
@@ -17,170 +17,410 @@
...
+        Args:
+            stream: Parameter not used.
+            instruction: The Qiskit instruction to translate to cQASM.
+            binary_control: Parameter not used.
+
+        Raises:
+            ApiError: the gate is not supported by the circuit parser.
+
+        """
+        del stream  # unused
```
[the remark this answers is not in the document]
QFer: The signature of the method has to be the same as an existing gate to call it via the construction:
            gate_function = getattr(self, gate_name, getattr(self, "_gate_not_supported"))
            gate_function(stream, instruction)
in the parse methods. To prevent a warning from static code inspection I del't the unused parameters.

src/quantuminspire/qiskit/circuit_parser.py
```diff
@@ -17,170 +17,410 @@
...
         """ Translates the controlled X element.
 
         Args:
-            circuit: The Qiskit circuit with CX element.
+            stream: The string-io stream to where the resulting cQASM is written.
+            instruction: The Qiskit instruction to translate to cQASM.
+
+        """
+        qubit_indices = tuple(instruction.qubits)
+        stream.write('CNOT q[{0}], q[{1}]\n'.format(*qubit_indices))
```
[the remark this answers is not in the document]
QFer: It was done this way in the original code. I never questioned it. Changed it.

src/quantuminspire/qiskit/circuit_parser.py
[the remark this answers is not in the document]
quantumkoen: I saw this comment in my email for some reason, and wanted to but in: I think `del`eting an unused parameter is a bit harsh. It is python convention to replace an unused parameter with a single underscore, `_`. This will shut up static code inspection and not destruct something of which you do not know if it is ok to destroy it.

src/quantuminspire/qiskit/circuit_parser.py
```diff
@@ -17,170 +17,410 @@
...
+            instruction: The Qiskit instruction to translate to cQASM.
+            axis: The axis for which the Rotation operator is parsed ('x', 'y' or 'z').
+            binary_control: The multi-bits control string. The gate is executed when all specified classical bits are 1.
+
+        """
+        angle_q0 = instruction.params[0]
+        qubit_indices = tuple(instruction.qubits)
+        stream.write('C-R{0} {1}q[{2}], {3:.6f}\n'.format(axis, binary_control, *qubit_indices, angle_q0))
+
+    def _rx(self, stream: StringIO, instruction: QobjItem) -> None:
```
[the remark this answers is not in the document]
QFer: Ok. I made them all static now, solving [DEM-542], but I doubt if this is better OO. There is in my opinion no need to make the members static because an instance of CircuitToString is available when the parse method is called.

src/quantuminspire/qiskit/circuit_parser.py
```diff
@@ -17,170 +17,410 @@
...
+        Args:
+            stream: The string-io stream to where the resulting cQASM is written.
+            instruction: The Qiskit instruction to translate to cQASM.
+            binary_control: The multi-bits control string. The gate is executed when all specified classical bits are 1.
+
+        """
+        self._c_u3(stream, instruction, binary_control)
+
+    @staticmethod
+    def _u0(stream: StringIO, instruction: QobjItem) -> None:
```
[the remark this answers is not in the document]
QFer: I existed already (also not used). When we don't add u0 as a basis_gate to the configuration, this method is never called. So I deleted it for now.

src/quantuminspire/qiskit/circuit_parser.py
```diff
@@ -191,60 +431,135 @@ def _u3(self, circuit: Dict[str, Any]) -> str:
...
-        Returns:
-            cQASM code string.
         """
-        angle_q0 = circuit['params'][0]
-        qubit_indices = tuple(circuit['qubits'])
-        # return 'R%c q[%d], %f\n' % (axis, *qubit_indices, angle_q0)
-        return 'R{0} q[{1}], {2:f}\n'.format(axis, *qubit_indices, angle_q0)
+        gates = ['C-Rz', 'C-Ry', 'C-Rz']
+        binary_controls = [binary_control] * 3
+        angles = list(instruction.params[i] for i in [2, 0, 1])
```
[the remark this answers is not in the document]
QFer: Only the preparation of angles and index for the write is the same. I think a common method that can handle both the u3 and its controlled counterpart would be more complex to maintain than this.

src/quantuminspire/qiskit/circuit_parser.py
QFer: binary control in QX is always on all classical bits, therefore starting with 0.
In qiskit you have something like gate.c_if(c, 6) where the integer value actually means some bits must be 1 and others must be 0. Because qx can only check bits being true, we have to inverse the bits that has to be zero and check them all for being 1, so starting from 0:n for n+1 classical bits.
I know b[0:0] generates a syntax error from the web interface but I think that is an omission in the web interface. I first had the check for a one bit program, but deleted it later, because it is legal and one conditional path to maintain less. The spec does say nothing about i < j for the construction b[i:j]. There are a few other constructions that the web interface cannot handle (for example C-Rx b[0], q[0], 3.14, also all controlled multi qubit gates are not parsed correctly by the web interface) but QX is compatible with the spec.

src/tests/quantuminspire/qiskit/test_circuit_parser.py
QFer: Oke, I'll fix this. I agree it is better mainainable when they are tested in seperated tests.

src/tests/quantuminspire/qiskit/test_circuit_parser.py
QFer: :) 👍
This bla-gate will swing the qubit to higher energy levels. But then we have the 'blabla' gate to test the non-existent gate and have to come up with another gate etc.

src/quantuminspire/qiskit/circuit_parser.py
quantumkoen: > binary control in QX is always on all classical bits, therefore starting with 0.

This is incorrect. Constructions like `c-X b[3], b[5], q[2]` are perfectly legal in cqasm (and thus in qx).

Also, if you negate classical bits for a particular c-? instruction, you should negate them back to their original value in case those same bits are used again later.

QFer: I meant sent to QX as in constructed by the SDK. Changed it.

QFer: In qiskit the gate:
`X(q[2]).c_if(c, 40)`
cannot be simply translated in:
`c-X b[3], b[5], q[2]`
because int(40) is bin(101000). The zero bits must be checked too. in cQASM classical bits can only be check being positive. That's why we have to negate the zero-bits before and after the conditional gate and checking all the classical bits as:
`not b[zero bits]
c-X b[3], b[5], q[2]
not b[zero bits]`

src/quantuminspire/qiskit/circuit_parser.py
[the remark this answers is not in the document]
QFer: ok, good alternative I didn't think of.

PR #45, opened 2019-04-11
[DEM-851] Mybinder support

QFer:
* Mybinder support for our example jupyter notebooks


environment.yml
```diff
@@ -0,0 +1,13 @@
...
+  - conda-forge
+dependencies:
+  - pip
+  - numpy==1.16.3
+  - pytest==4.4.1
+  - matplotlib>=2.1
+  - pip:
+    - coreapi==2.3.3
+    - quantuminspire==0.7.0
+    - qiskit==0.8.1
```
qSaevar: does this now mean that we have to maintain version of 3rd party libraries in two places? I.e., setup.py and environment.py? Mybinder claims to support requirements.txt and setup.py, can we not use that?

README.md
```diff
@@ -81,6 +81,11 @@ cd docs
 python example_projectq_grover.py
 ```
 
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/QuTech-Delft/quantuminspire.git/master?filepath=%2Fdocs)
+
+Another way to browse and run the available notebooks is by clicking the 'launch binder' button above.
+If you are experiencing delays, please check the 'Binder' system status. The service may be down.
```
qSaevar: What, or where, is the binder system status? this one: https://grafana.mybinder.org/d/3SpLQinmk/1-overview?refresh=1m&orgId=1 ?

QFer: I'll remove this line about system status.

environment.yml
QFer: I'll change it to use a requirements.txt. I'll put the dependencies needed for DEM850 also in.

PR #46, opened 2019-04-12
[DEM-870] Mypy in travis

QFer:
* matplotlib and jupyter both dependencies in setup instead of separate install because the notebooks need them to run.
* install mypy
* run mypy
* Increased time-out of the notebook tests because of the performance of the staging-environment.


QFer:
Some tests are depending on the performance of the staging environment. I've increased the time-out of the notebooks to 5 minutes. When the time out was 3 minutes and the integration tests on staging were running at the same time, we got a time-out.

PR #48, opened 2019-04-12
[DEM-867] QIBackend creation with all defaults

QFer:
* No api needed when creating QIBackend.
* In QIBackend, try to construct default QuantumInspireApi when no Api is passed to the constructor.
* When something goes wrong, overrule the technical message with a meaningful message what to do.
* Unit tests adjusted.

### DIFF
--- a/src/quantuminspire/projectq/backend_qx.py
+++ b/src/quantuminspire/projectq/backend_qx.py
@@ -30,7 +30,7 @@ from projectq.ops import (NOT, Allocate, Barrier, Deallocate, FlushGate, H,
                           Measure, Ph, Rx, Ry, Rz, S, Sdag, Swap, T, Tdag, X,
                           Y, Z, Command)
 from projectq.types import Qubit
-from quantuminspire.exceptions import ApiError
+from quantuminspire.exceptions import AuthenticationError
 from quantuminspire.api import QuantumInspireAPI
 from quantuminspire.exceptions import ProjectQBackendError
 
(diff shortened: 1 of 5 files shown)


src/quantuminspire/projectq/backend_qx.py
```diff
@@ -63,8 +64,12 @@ def __init__(self, num_runs: int = 1024, verbose: int = 0, quantum_inspire_api:
         self._measured_ids: List[int] = []
         self._allocated_qubits: Set[int] = set()
         self._max_qubit_id: int = -1
-        if not quantum_inspire_api:
-            raise RuntimeError("Api is required")
+        if quantum_inspire_api is None:
+            try:
+                quantum_inspire_api = QuantumInspireAPI()
+            except ApiError as ex:
+                raise ApiError('Make sure you have saved your token credentials on disk (see README.md) or '
```
quantumkoen: If you `pip install` this from the cheeseshop, will you be able to find README.md ?

src/quantuminspire/projectq/backend_qx.py
```diff
@@ -63,8 +64,12 @@ def __init__(self, num_runs: int = 1024, verbose: int = 0, quantum_inspire_api:
         self._measured_ids: List[int] = []
         self._allocated_qubits: Set[int] = set()
         self._max_qubit_id: int = -1
-        if not quantum_inspire_api:
-            raise RuntimeError("Api is required")
+        if quantum_inspire_api is None:
+            try:
+                quantum_inspire_api = QuantumInspireAPI()
+            except ApiError as ex:
```
quantumkoen: According to the docstring on `QuantumInspireAPI`, `ApiError` is raised either when authentication is not available or when the schema could not be loaded. The reraise below suggests to the user that the error is always that there are no token credentials saved. That seems inconsistent to me. This means that if the reason for the `ApiError` is not that there is no authentication, the user will think they need to specify authentication while they already have, greatly confusing the user.

Perhaps this means the exceptions in the `QuantumInspireAPI` shoudl not be of the same class, and a new class should be introduced, `AuthenticationError` or something like that?

src/quantuminspire/projectq/backend_qx.py
QFer: Nope, I'll remove that reference to README.md. I assume that someone who is going to use quantuminspire SDK can find this information op pypi (or github).

src/quantuminspire/projectq/backend_qx.py
QFer: Agree, I'll add AuthenticationError for this.

PR #49, opened 2019-05-02
[DEM-NNN] Limiting qiskit version

QFer:
* For compatibility reasons, the version of qiskit is limited to 0.8.1


QFer:
Quantuminspire version bumped to 0.7.0

PR #53, opened 2019-05-10
[DEM-959] Validate usage of classical bits after measurement

QFer:
* Adjusted validation method in qiskit backend_qx (__validate_no_usage_after_measure)
  * docstring was not correct
  * added a check that measured memory bits are not used later on as a condition in a binary controlled gate. Because in FSP simulation measurements are not done until the end, so we can't make usage of measured classical bits to influence the outcome of a circuit yet.
* Unittests added

### DIFF
--- a/src/quantuminspire/qiskit/backend_qx.py
+++ b/src/quantuminspire/qiskit/backend_qx.py
@@ -207,7 +207,7 @@ class QuantumInspireBackend(BaseBackend):  # type: ignore
 
         for experiment in job.experiments:
             self.__validate_number_of_clbits(experiment)
-            QuantumInspireBackend.__validate_no_gates_after_measure(experiment)
+            QuantumInspireBackend.__validate_no_usage_after_measure(experiment)
 
     @staticmethod
     def __validate_number_of_shots(job: QasmQobj) -> None:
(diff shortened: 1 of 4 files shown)


src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -264,6 +270,19 @@ def __validate_no_gates_after_measure(experiment: QasmQobjExperiment) -> None:
                     elif qubit in measured_qubits:
                         raise QisKitBackendError('Operation on qubit {} after measurement'.format(qubit))
 
+        memory_bits = []
+        for instruction in experiment.instructions:
```
qSaevar: This is the same loop as above, is it not enough to loop once and check for  

>             1. When qubits are used again after being measured.
>             2. When conditional gate operations (binary controlled gates) are done based on previously measured
>                classical bits

in the same loop?

QFer: My idea to keep it seperated was that the second block could be deleted when we are also doing non-FSP (another ticket).
I coded now one loop.

qSaevar: I see your point, now this whole loop can be bypassed when non-fsp is supported as reuse of qubits after measure will also be allowed. So now, or then, this loop can be extracted into a method that is only called when fsp is used.

QFer:
I tested the notebooks also. They were good.

PR #55, opened 2019-05-20
[DEM-851] Mybinder revise

QFer:
* Requirements.txt is needed for mybinder to install projectq and Qiskit.
* Adjusted requirements.txt. Only projectq and qiskit dependencies without version numbers. When merged in master Mybinder will build the master branch against the latest version of qiskit/projectq.


QFer:
I had to rebase before merging because of other pull request

PR #58, opened 2019-06-19
[DEM-944] fsp vs non-fsp in qiskit

QFer:
The following general rules are implemented for both Qiskit and ProjectQ (another PR):

* An algorithm without measurements or only measurements at the end (no gates following the measurements) can be executed as full state projection, otherwise it is executed as non-fsp.
* Also in a non-fsp algorithm not more than one different qubits may be measured to a classical bit and a qubit may not be measured to more than one classical bit.
* When a non-fsp algorithm is run the warning as specced in the desired outcome of the ticket [DEM-944] should be printed, according to the acceptance criteria. In api there is a method to enable or disable this warning:

show_fsp_warning(self, enable: bool = True)

[DEM-944]: https://qutech-sd.atlassian.net/browse/DEM-944

### DIFF
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -473,7 +473,7 @@ class QuantumInspireAPI:
         }
         if not full_state_projection and self.enable_fsp_warning:
             logger.warning("Your experiment can not be optimized and may take longer to execute, "
-                          "see https://www.quantum-inspire.com/kbase/optimization-of-simulations/ for details.")
+                           "see https://www.quantum-inspire.com/kbase/optimization-of-simulations/ for details.")
         return OrderedDict(self._action(['jobs', 'create'], params=payload))
 
     #  results  #
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -16,7 +16,7 @@ limitations under the License.
 """
 
 import itertools
-import warnings
+import logging
 import time
 import uuid
 from typing import Type, List, Dict, Union, Optional, Any
@@ -31,6 +31,7 @@ from quantuminspire.exceptions import ApiError, AuthenticationError
 from quantuminspire.job import QuantumInspireJob
 
 QI_URL = 'https://api.quantum-inspire.com'
+logger = logging.getLogger(__name__)
 
 
 class QuantumInspireAPI:
@@ -115,7 +116,7 @@ class QuantumInspireAPI:
 
     def show_fsp_warning(self, enable: bool = True) -> None:
         """ The warning that is printed when a non-FSP (full state projection) job is about to run can be controlled,
-        i.e. enabled or disabled via this method.
+            i.e. enabled or disabled via this method.
 
         Args:
             enable: when True the fsp-warning is shown, otherwise not.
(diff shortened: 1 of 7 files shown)


QFer:
Rebased

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -248,41 +257,51 @@ def __validate_number_of_clbits(self, experiment: QasmQobjExperiment) -> None:
...
+        fsp = True
+        measurement_found = False
         for instruction in experiment.instructions:
             if instruction.name == 'measure':
-                for qubit in instruction.qubits:
-                    measured_qubits.append(qubit)
-                for qubit in instruction.memory:
-                    memory_bits.append(qubit)
+                measurement_found = True
             else:
```
qSaevar: why not `elif measurement_found:` instead of `else: if measurement_found: `?

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -300,89 +317,138 @@ def test_validate_nr_classical_qubits_less_than_nr_qubits_conditional_gate(self)
...
-            job_dict = self._basic_qobj_dictionary
-            job_dict['experiments'][0]['instructions'] = instructions
-            job = qiskit.qobj.Qobj.from_dict(job_dict)
-            self.assertRaises(QisKitBackendError, simulator.run, job)
-
-    def test_no_operation_after_measure_cx_gate(self):
-        with patch.object(QuantumInspireBackend, "_submit_experiment", return_value=Mock()):
-            simulator = QuantumInspireBackend(Mock(), Mock())
-            instructions = [{'name': 'X', 'qubits': [1]},
+    def test_for_full_state_projection(self):
```
qSaevar: Testing four cases in one unit test, should it not be four tests?

src/quantuminspire/api.py
```diff
@@ -458,6 +471,9 @@ def _create_job(self, name: str, asset: Dict[str, Any], project: Dict[str, Any],
             'full_state_projection': full_state_projection,
             'user_data': user_data
         }
+        if not full_state_projection and self.enable_fsp_warning:
+            logger.warning("Your experiment can not be optimized and may take longer to execute, "
+                          "see https://www.quantum-inspire.com/kbase/optimization-of-simulations/ for details.")
```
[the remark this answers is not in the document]
QFer: fixed

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -226,7 +224,18 @@ def __validate_number_of_shots(job: QasmQobj) -> None:
             raise QisKitBackendError('Invalid shots (number_of_shots={})'.format(number_of_shots))
 
     def __validate_number_of_clbits(self, experiment: QasmQobjExperiment) -> None:
-        """ Checks whether the number of classical bits has a valid value.
+        """ Checks whether the number of classical bits has a value cQASM can support.
+            1. When number of classical bits is less than 1 an error is raised.
```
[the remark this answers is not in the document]
QFer: ok

src/quantuminspire/qiskit/backend_qx.py
QFer: ok

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: I'll split them up

PR #59, opened 2019-06-20
Update images, dataplotter class

RobertWezemanTNO:
Update for images.

Restored the instance of the Data_plotter class. This resolves the problem of the images not being displayed next to each other.


QFer:
Updated in pull request: https://github.com/QuTech-Delft/quantuminspire/pull/61

PR #70, opened 2019-10-11
[DEM-1172] Endless waiting for job

QFer:
* At two places (api and qi_job) the loop is now also ended when other 'job final states' are received as status.
* Unit tests for these 'final state' status values added.

### DIFF
--- a/src/tests/quantuminspire/test_api.py
+++ b/src/tests/quantuminspire/test_api.py
@@ -1004,6 +1004,16 @@ class TestQuantumInspireAPI(TestCase):
         is_completed = api._wait_for_completed_job(quantum_inspire_job, collect_max_tries, sec_retry_delay=0.0)
         self.assertFalse(is_completed)
 
+    def test_wait_for_cancelled_job_returns_false(self):
+        job_id = 509
+        expected_payload = {'id': job_id}
+        self.coreapi_client.handlers['jobs'] = partial(self.__mock_job_handler, expected_payload, 'read',
+                                                       status='CANCELLED')
+        api = QuantumInspireAPI('FakeURL', self.authentication, coreapi_client_class=self.coreapi_client)
+        quantum_inspire_job = QuantumInspireJob(api, job_id)
+        is_completed = api._wait_for_completed_job(quantum_inspire_job)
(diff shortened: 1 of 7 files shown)


src/tests/quantuminspire/test_api.py
```diff
@@ -1004,6 +1004,16 @@ def test_wait_for_completed_job_returns_false(self):
...
         self.assertFalse(is_completed)
 
+    def test_wait_for_cancelled_job_returns_false(self):
+        job_id = 509
+        expected_payload = {'id': job_id}
+        self.coreapi_client.handlers['jobs'] = partial(self.__mock_job_handler, expected_payload, 'read',
+                                                       status='CANCELLED')
+        api = QuantumInspireAPI('FakeURL', self.authentication, coreapi_client_class=self.coreapi_client)
+        quantum_inspire_job = QuantumInspireJob(api, job_id)
+        is_completed = api._wait_for_completed_job(quantum_inspire_job)
```
qSaevar: Not a big fan of calling private methods but I see it is done a lot in this test class.

src/tests/quantuminspire/test_api.py
```diff
@@ -1126,6 +1142,14 @@ def __mocks_for_api_execution(self):
 
         return expected_job_result, job_mock, asset_mock, backend_mock, project_mock
 
+    def test_execute_qasm_cancelled_job(self):
+        mocks = self.__mocks_for_api_execution(fake_no_results=True)
+        api = QuantumInspireAPI('FakeURL', self.authentication, coreapi_client_class=self.coreapi_client)
+        qasm = 'version 1.0...'
+        number_of_shots = 1024
+        results = api.execute_qasm(qasm, number_of_shots=number_of_shots, backend_type=1)
+        self.assertEqual(results, OrderedDict())
```
qSaevar: How does the user know if the job was cancelled, failed, or just timed out?

QFer: When the result was False, now a minimal result object is created with an empty histogram and an error message. so in all cases some result is sent back and the user can check the result for more information.

PR #74, opened 2019-11-19
[DEM-1194] Full state projection

QFer:
* Full state projection defaults to False for algorithms executed directly (without projectQ/Qiskit backend)
* Added some missing license headers
* Reordered some imports
* Solved python code inspection warning

### DIFF
--- a/src/quantuminspire/qiskit/quantum_inspire_provider.py
+++ b/src/quantuminspire/qiskit/quantum_inspire_provider.py
@@ -20,10 +20,10 @@ from typing import List, Optional, Any
 import coreapi
 from qiskit.providers import BaseProvider
 
-from quantuminspire.credentials import get_token_authentication, get_basic_authentication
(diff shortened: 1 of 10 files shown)


src/quantuminspire/qiskit/quantum_inspire_provider.py
```diff
@@ -60,6 +76,7 @@ def set_authentication_details(self, email: str, password: str, qi_url: str = QI
             qi_url: URL that points to quantum-inspire api. Default value: 'https://api.quantum-inspire.com'.
 
         """
+        del qi_url
```
qSaevar: why do you delete the parameter instead of passing it along to `set_basic_authentication` ?

QFer: I resolved an unused parameter warning without changing the impact, but I think you are right. This should be in the call to 'set_basic_authentication'

PR #76, opened 2020-04-06
[DEM-1185] Result generates nice error when creating job fails

QFer:
* When create job fails or another error (e.g. from coreapi) occures when executing qasm, instead of passing the coreapi exception to the user we package it in the result (no histogram, raw text explaining what happened)
* Added unit tests

### DIFF
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -91,7 +91,7 @@ class QuantumInspireAPI:
         try:
             self._load_schema()
         except (CoreAPIException, TypeError) as ex:
-            raise ApiError('Could not connect to {}'.format(base_uri)) from ex
+            raise ApiError(f'Could not connect to {base_uri}') from ex
 
     def _get(self, uri_path: str) -> Any:
         """ Method for making requests to the coreapi client instance to get some piece of information. The information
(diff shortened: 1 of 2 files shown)


src/quantuminspire/api.py
```diff
@@ -475,7 +475,10 @@ def _create_job(self, name: str, asset: Dict[str, Any], project: Dict[str, Any],
         if not full_state_projection and self.enable_fsp_warning:
             logger.warning("Your experiment can not be optimized and may take longer to execute, "
                            "see https://www.quantum-inspire.com/kbase/optimization-of-simulations/ for details.")
-        return OrderedDict(self._action(['jobs', 'create'], params=payload))
+        try:
+            return OrderedDict(self._action(['jobs', 'create'], params=payload))
+        except (CoreAPIException, TypeError, ValueError) as err_msg:
+            raise ApiError('Job with name {} not created: {}'.format(name, err_msg)) from err_msg
```
oschusler, line 481: we try to move to f-strings as much as possible. Could you update these?

src/quantuminspire/api.py
```diff
@@ -835,6 +838,9 @@ def execute_qasm(self, qasm: str, backend_type: Optional[Union[Dict[str, Any], i
             has_results, message = self._wait_for_completed_job(quantum_inspire_job, collect_tries)
             return OrderedDict(quantum_inspire_job.retrieve_results()) if has_results else \
                 OrderedDict(self._generate_error_result(message))
+        except (CoreAPIException, TypeError, ValueError, ApiError) as err_msg:
+            message = 'Error raised while executing qasm: {}'.format(err_msg)
```
oschusler, line 842: and these

src/quantuminspire/api.py
QFer: Did them all in this file

src/quantuminspire/api.py
QFer: Did them all in this file

PR #77, opened 2020-04-09
[DEM-1418] No message about non-optimized experiments on hardware backends

QFer:
* Check if we are running on a hardware backend before printing the non-optimization message.
* Test added that no message is printed on hardware backend.

### DIFF
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -473,7 +473,7 @@ class QuantumInspireAPI:
             'full_state_projection': full_state_projection,
             'user_data': user_data
         }
-        if not full_state_projection and self.enable_fsp_warning and not backend_type["is_hardware_backend"]:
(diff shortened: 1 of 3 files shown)


src/quantuminspire/api.py
```diff
@@ -472,7 +473,7 @@ def _create_job(self, name: str, asset: Dict[str, Any], project: Dict[str, Any],
             'full_state_projection': full_state_projection,
             'user_data': user_data
         }
-        if not full_state_projection and self.enable_fsp_warning:
+        if not full_state_projection and self.enable_fsp_warning and not backend_type["is_hardware_backend"]:
```
oschusler, line 476:
```suggestion
        if not full_state_projection and self.enable_fsp_warning and not backend_type.get("is_hardware_backend", False):
```

QFer: Looks good. It's safer the way you propose.

PR #81, opened 2020-04-23
[DEM-1447] updated the descriptions of the api structures

QFer:
* updated the docstring-descriptions of the api structures in some methods

### DIFF
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -680,7 +680,8 @@ class QuantumInspireAPI:
                 | contentType (str)         | The description of the content e.g. 'application/qasm' or
                 |                           | 'text/plain'.
                 | content (str)             | The content itself. For example a cQASM program when linked to a job.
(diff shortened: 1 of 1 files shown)


src/quantuminspire/api.py
```diff
@@ -665,6 +680,7 @@ def get_asset(self, asset_id: int) -> Dict[str, Any]:
                 | contentType (str)         | The description of the content e.g. 'application/qasm' or
                 |                           | 'text/plain'.
                 | content (str)             | The content itself. For example a cQASM program when linked to a job.
+                | measurement_mask (int)    | A mask for the measured bits in the cQASM program.
```
rhjdjong, line 683: Maybe specify here that this is a read-only parameter that gets calculated when the asset is assigned to a job.

QFer: All parameters are read only in SDK. I will add the extra remark when it is calculated

PR #85, opened 2020-05-11
[DEM-1442] Supporting hardware backends

QFer:
* Using backend info to limit or enhance functionality as allowed_gates/max nr. of shots etc for Qiskit and projectQ backends
* Added support for cz gates in Qiskit /ProjectQ backend
* Added support for cr gate in ProjectQ backend.


docs/example_projectq_grover.py
```diff
@@ -111,10 +111,11 @@ def get_authentication():
     # Remote Quantum-Inspire backend #
     authentication = get_authentication()
     qi = QuantumInspireAPI(QI_URL, authentication)
+    qi_backend = QIBackend(quantum_inspire_api=qi)
 
-    compiler_engines = restrictedgateset.get_engine_list(one_qubit_gates="any", two_qubit_gates=(CNOT, CZ, Toffoli))
+    compiler_engines = restrictedgateset.get_engine_list(one_qubit_gates=qi_backend.one_qubit_gates,
```
oschusler, line 116: Given that toffoli is not a two_qubit_gate, I understand why you remove it here, but how do we add three_qubit_gates?

src/quantuminspire/api.py
```diff
@@ -812,8 +812,8 @@ def _generate_error_result(message: str) -> Dict[str, Any]:
         return result_obj
 
     def execute_qasm(self, qasm: str, backend_type: Optional[Union[Dict[str, Any], int, str]] = None,
-                     number_of_shots: int = 256, collect_tries: Optional[int] = None,
-                     default_number_of_shots: int = 256, identifier: Optional[str] = None,
+                     number_of_shots: int = 1024, collect_tries: Optional[int] = None,
```
oschusler, line 815: Shouldn't this be None, with a check that if `number_of_shots` is None, it should be retrieved from the backend_type object?

src/quantuminspire/projectq/backend_qx.py
```diff
@@ -303,7 +395,8 @@ def _store(self, cmd: Command) -> None:
             self._measured_ids += [logical_id]
             # do not add the measurement statement when fsp is possible
             if not self._full_state_projection:
-                self.qasm += "\nmeasure q[{}]".format(sim_qubit_id)
+                if self._is_simulation_backend:
+                    self.qasm += "\nmeasure q[{}]".format(sim_qubit_id)
```
oschusler, line 399: Can you make this an f-string?

src/quantuminspire/projectq/backend_qx.py
```diff
@@ -483,7 +576,7 @@ def _execute_cqasm(self) -> None:
         self._quantum_inspire_result = self._quantum_inspire_api.execute_qasm(
             self._cqasm,
             number_of_shots=self._num_runs,
-            backend_type=self._backend_type,
+            backend_type=self._backend,
```
oschusler, line 579: why did you name this backend? It should be a backend_type.

src/quantuminspire/projectq/backend_qx.py
```diff
@@ -58,27 +61,106 @@ def __init__(self, num_runs: int = 1024, verbose: int = 0, quantum_inspire_api:
...
+        self._two_qubit_gates: Tuple[Any, ...] = self._get_two_qubit_gates()
+        self._three_qubit_gates: Tuple[Any, ...] = self._get_three_qubit_gates()
+
+    def _get_one_qubit_gates(self) -> Tuple[Any, ...]:
+        if len(self._backend['allowed_operations']) > 0:
+            one_qubit_gates = []
+            for keys in self._backend['allowed_operations']:
+                if keys in ['single_gates', 'parameterized_single_gates']:
+                    for gate in self._backend['allowed_operations'][keys]:
+                        if gate == 'x':
```
oschusler, line 94: Can't we do this smarter? All names have the first character capitalized, and can be imported via importlib?

src/quantuminspire/qiskit/circuit_parser.py
```diff
@@ -47,6 +47,29 @@ def _gate_not_supported(_stream: StringIO, instruction: QasmQobjInstruction, _bi
...
+    def _c_cz(stream: StringIO, instruction: QasmQobjInstruction, binary_control: str) -> None:
+        """ Translates the binary-controlled controlled Z element.
+
+        Args:
+            stream: The string-io stream to where the resulting cQASM is written.
+            instruction: The Qiskit instruction to translate to cQASM.
+            binary_control: The multi-bits control string. The gate is executed when all specified classical bits are 1.
+
+        """
+        stream.write('C-CZ {0}q[{1}], q[{2}]\n'.format(binary_control, *instruction.qubits))
```
oschusler, line 71: Could you make this an f-string, and don't you miss a space between `{0}` and `q`?

docs/example_projectq_grover.py
QFer: It was an error (in my opinion) to add toffoli to the 2-qubit gates.
I did the following:
- Besides the one_qubit_gates and two_qubit_gates there is a possibility to add 'other gates to the restrictedgateset engine. This parameter is called other_gates. I will add the three qubit gates to this parameter (thus the toffoli). Because the old example was actually without toffoli (because it wasn't a 2 bit gate) I added only the one and two qubit gates. I will for this example add the three qubit gates also.

src/quantuminspire/api.py
QFer: You are right. Nice suggestion

src/quantuminspire/projectq/backend_qx.py
QFer: I changed all strings to f strings in this file.

src/quantuminspire/projectq/backend_qx.py
QFer: I changed it back

src/quantuminspire/qiskit/circuit_parser.py
QFer: I'll do it for the whole file then. The space is in the binary_control string

QFer: I think the 'old type' string is more clear than f string. Maybe you could look at method _cz and _cx and check yourself. There is no need to go to f-strings, is there?

src/quantuminspire/projectq/backend_qx.py
QFer: I use getattr(sys.modules[__name__] because all operation classes are imported from projectq locally. No need to import from some other module.

src/quantuminspire/projectq/backend_qx.py
```diff
@@ -58,27 +61,82 @@ def __init__(self, num_runs: int = 1024, verbose: int = 0, quantum_inspire_api:
...
+        else:
+            one_qubit_gates = [X, Y, Z, H, S, Sdag, T, Tdag, Rx, Ry, Rz]
+        return tuple(one_qubit_gates)
+
+    def _get_two_qubit_gates(self) -> Tuple[Any, ...]:
+        if len(self._backend_type['allowed_operations']) > 0:
+            two_qubit_gates = []
+            for keys in self._backend_type['allowed_operations']:
+                if keys in ['dual_gates', 'parameterized_dual_gates']:
+                    for gate in self._backend_type['allowed_operations'][keys]:
```
oschusler, line 105: I would swap the two. Loop over the two gate_sets that we are interested in, and check whether they are present.
```suggestion
        allowed_operations = self._backend_type['allowed_operations']
        if len(allowed_operations) > 0:
            two_qubit_gates = []
            for gate_set in ['dual_gates', 'parameterized_dual_gates']:
                if gate_set in allowed_operations['allowed_operations']:
                    for gate in allowed_operations[gate_set]:
```

oschusler: btw, same for the `_get_one_qubit_gates` and `_get_three_qubit_gates`

src/quantuminspire/qiskit/circuit_parser.py
oschusler: There is not necessarily, expect that Python prescribes the use of it in their style guide.

src/quantuminspire/projectq/backend_qx.py
QFer: Done

PR #86, opened 2020-07-09
[DEM-1494] Changes for development version of Qiskit Terra (0.15.0)

QFer:
* The probabilities histogram was added to the Qiskit result data. The Qiskit-Terra changes now don't copy these extra fields into the result object.
* The import 'qiskit.validation.base' is now invalid, needed for Obj class.
* Added a QIResult, subclassed from Result and added handling for probabilities result.


QFer:
We should only merge these changes when Qiskit updated to pypi and we re-tested this solution!

oschusler (approved):
Not entirely sure why we suddenly have a QI specific result, but I assume that the generic Result format of Qiskit changed, and this is more tailored to us?

QFer:
> 
> 
> Not entirely sure why we suddenly have a QI specific result, but I assume that the generic Result format of Qiskit changed, and this is more tailored to us?

Earlier it was possible to insert our probabilities and counts results in the Qiskit Result structure (I discussed the best place for it with someone from Qiskit). The validations on this structure, added by qiskit, were only on their own fields, not on ours. In the end we had our results in the Result structure.
This was changed. The Result structure is now a class with strict validations so our fields were not copied through anymore in the end result.
To solve this I subclassed the Result of Qiskit (qi_result) and added our results to this subclass. Now our results are not mixed up anymore. I had to add some functionality to generate a nice probabilities result for the end user.

PR #87, opened 2020-09-06
[DEM-NNN] Preparations for release 1.3.0

QFer:
* Version update
* Ipython notebook updates


oschusler (approved):
Everytime we update the example notebooks, there are huge changes due to the embedded images. Is there anything we can do about that?

PR #92, opened 2021-02-05
[INS-126] Project names in Qiskit

QFer:
* When using Qiskit, now also Quantum Inspire project names can be set
* When no project name is set, a default is taken
* When an existing project name is chosen, the experiments are added to the existing project
* qi_job.result() now has a parameter to only fetch the results for the latest run (instead ALL results for the project)

### DIFF
--- a/src/quantuminspire/projectq/backend_qx.py
+++ b/src/quantuminspire/projectq/backend_qx.py
@@ -58,7 +58,7 @@ class QIBackend(BasicEngine):  # type: ignore
         """ Because engines are meant to be 'single use' by the way ProjectQ is designed,
         any additional gates received after a FlushGate triggers an exception. """
         self._clear: bool = True
-        self.qasm: str = ""
+        self._qasm: str = ""
         self._reset()
         self._verbose: int = verbose
         self._cqasm: str = str()
@@ -145,6 +145,11 @@ class QIBackend(BasicEngine):  # type: ignore
         """ Return cqasm code that is generated last. """
         return self._cqasm
 
+    @property
+    def qasm(self) -> str:
+        """ Return qasm code at any moment in the process. """
+        return self._qasm
+
     def is_available(self, cmd: Command) -> bool:
         """
         Via this method the ProjectQ framework determines which commands (gates) are available in the backend.
@@ -190,7 +195,7 @@ class QIBackend(BasicEngine):  # type: ignore
         """ Reset temporary variable qasm to an initial value and set a flag to clear variables in _store
             when _store is called. """
         self._clear = True
-        self.qasm = ""
+        self._qasm = ""
 
     def _allocate_qubit(self, index_to_add: int) -> None:
         """ On a simulation backend it is possible to reuse qubits. The advantage of reusing qubits is that less
@@ -262,8 +267,8 @@ class QIBackend(BasicEngine):  # type: ignore
 
                     # to reuse a de-allocated bit we do a prep_z first, which is better implemented as a
                     # measurement and binary controlled x-gate
-                    self.qasm += f"\nmeasure q[{allocation_entry[0]}]"
-                    self.qasm += f"\nc-x b[{allocation_entry[0]}], q[{allocation_entry[0]}]"
+                    self._qasm += f"\nmeasure q[{allocation_entry[0]}]"
+                    self._qasm += f"\nc-x b[{allocation_entry[0]}], q[{allocation_entry[0]}]"
                     index = self._allocation_map.index(allocation_entry)
                     self._allocation_map[index] = (allocation_entry[0], index_to_add)
 
@@ -311,8 +316,8 @@ class QIBackend(BasicEngine):  # type: ignore
                 raise RuntimeError(f"Bit position in simulation backend not found for"
                                    f" physical bit {physical_qubit_id}.")
             return allocation_entry[0]
-        else:
-            return physical_qubit_id
+
+        return physical_qubit_id
 
     def _switch_fsp_to_nonfsp(self) -> None:
         """ We have determined that the algorithm is non-deterministic and cannot use fsp.
@@ -322,7 +327,7 @@ class QIBackend(BasicEngine):  # type: ignore
         for logical_qubit_id in self._measured_ids:
             physical_qubit_id = self._logical_to_physical(logical_qubit_id)
             sim_qubit_id = self._physical_to_simulated(physical_qubit_id)
-            self.qasm += f"\nmeasure q[{sim_qubit_id}]"
+            self._qasm += f"\nmeasure q[{sim_qubit_id}]"
         self._full_state_projection = False
 
     def _store(self, cmd: Command) -> None:
@@ -339,7 +344,7 @@ class QIBackend(BasicEngine):  # type: ignore
 
         if self._clear:
             self._clear = False
-            self.qasm = ""
+            self._qasm = ""
             self._measured_states = {}
             self._measured_ids = []
             self._full_state_projection = not self._backend_type["is_hardware_backend"]
@@ -374,7 +379,7 @@ class QIBackend(BasicEngine):  # type: ignore
             # do not add the measurement statement when fsp is possible
             if not self._full_state_projection:
                 if self._is_simulation_backend:
-                    self.qasm += f"\nmeasure q[{sim_qubit_id}]"
+                    self._qasm += f"\nmeasure q[{sim_qubit_id}]"
             return
 
         # when we find a gate after measurements we don't have fsp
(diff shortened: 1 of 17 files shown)


src/quantuminspire/projectq/backend_qx.py
```diff
@@ -60,13 +58,15 @@ def __init__(self, num_runs: int = 1024, verbose: int = 0, quantum_inspire_api:
         """ Because engines are meant to be 'single use' by the way ProjectQ is designed,
         any additional gates received after a FlushGate triggers an exception. """
         self._clear: bool = True
+        self.qasm: str = ""
```
rhjdjong, line 61: Is `qasm` part of the public interface of QIBackend? If not, it should be prefixed with an underscore.

src/tests/quantuminspire/qiskit/test_quantum_inspire_provider.py
```diff
@@ -210,10 +210,23 @@ def test_set_authentication(self):
...
             authentication = BasicAuthentication('email', 'password')
             quantum_inpire_provider.set_authentication(authentication)
-            api.assert_called_with(QI_URL, authentication)
+            api.assert_called_with(QI_URL, authentication, None)
+
+    def test_set_projectname(self):
+        with mock.patch('quantuminspire.qiskit.quantum_inspire_provider.QuantumInspireAPI') as api:
+            quantum_inpire_provider = QuantumInspireProvider()
+            token = 'This_is_a_nice_looking_token'
+            token = 'This_is_a_nice_looking_token'
```
rhjdjong, line 222: Remove duplicated line

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -163,10 +165,10 @@ def test_get_experiment_results_raises_simulation_error_when_no_histogram(self):
         job.job_id.return_value = '42'
         simulator = QuantumInspireBackend(api, Mock())
         with self.assertRaises(QisKitBackendError) as error:
-            simulator.get_experiment_results(job)
+            simulator.get_experiment_results(job, False)
```
rhjdjong, line 168: The meaning of the second argument is not clear. Consider using the explicit parameter name: `only_latest_run=False`

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -186,7 +188,7 @@ def test_get_experiment_results_returns_correct_value(self):
         api.get_jobs_from_project.return_value = [jobs]
         job = QIJob('backend', '42', api)
         simulator = QuantumInspireBackend(api, Mock())
-        experiment_result = simulator.get_experiment_results(job)[0]
+        experiment_result = simulator.get_experiment_results(job, False)[0]
```
rhjdjong, line 191: The meaning of the second argument is not clear. Consider using the explicit parameter name: only_latest_run=False

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -197,6 +199,67 @@ def test_get_experiment_results_returns_correct_value(self):
...
+        job['user_data'] = json.dumps(user_data)
+
+        api.get_job.side_effect = [job]
+        quantuminspire_job = Mock()
+        quantuminspire_job.get_job_identifier.side_effect = [1]
+        qijob = QIJob('backend', '42', api)
+        qijob.add_job(quantuminspire_job)
+
+        simulator = QuantumInspireBackend(api, Mock())
+        experiment_result = simulator.get_experiment_results(qijob, True)[0]
```
rhjdjong, line 227: The meaning of the second argument is not clear. Consider using the explicit parameter name: `only_latest_run=True`

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -219,7 +282,7 @@ def test_get_experiment_results_returns_single_shot(self):
         api.get_jobs_from_project.return_value = [jobs]
         job = QIJob('backend', '42', api)
         simulator = QuantumInspireBackend(api, Mock())
-        experiment_result = simulator.get_experiment_results(job)[0]
+        experiment_result = simulator.get_experiment_results(job, False)[0]
```
rhjdjong, line 285: The meaning of the second argument is not clear. Consider using the explicit parameter name: `only_latest_run=False`

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -259,7 +322,7 @@ def test_get_experiment_results_multiple_single_shots(self):
             api.get_jobs_from_project.return_value = [jobs]
             job = QIJob('backend', '42', api)
             simulator = QuantumInspireBackend(api, Mock())
-            experiment_result = simulator.get_experiment_results(job)[0]
+            experiment_result = simulator.get_experiment_results(job, False)[0]
```
rhjdjong, line 325: The meaning of the second argument is not clear. Consider using the explicit parameter name: `only_latest_run=False`

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -541,7 +596,7 @@ def run_histogram_test(self, single_experiment, mock_result1, mock_result2, expe
         self.mock_api.get_jobs_from_project.return_value = [jobs]
         job = QIJob('backend', '42', self.mock_api)
 
-        result = self.simulator.get_experiment_results(job)
+        result = self.simulator.get_experiment_results(job, False)
```
rhjdjong, line 599: The meaning of the second argument is not clear. Consider using the explicit parameter name: `only_latest_run=False`

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -171,25 +177,31 @@ def _submit_experiment(self, experiment: QasmQobjExperiment, number_of_shots: in
...
-
-    def get_experiment_results(self, qi_job: QIJob) -> List[ExperimentResult]:
+        quantum_inspire_job = self.__api.execute_qasm_async(compiled_qasm, backend_type=self.__backend,
+                                                            number_of_shots=number_of_shots, project=project,
+                                                            job_name=experiment.header.name,
+                                                            user_data=json.dumps(user_data),
+                                                            full_state_projection=full_state_projection)
+        return quantum_inspire_job
+
+    def get_experiment_results(self, qi_job: QIJob, only_latest_run: bool = True) -> List[ExperimentResult]:
```
rhjdjong, line 187: As you can see from the comments on the test scripts, the meaning of the second argument is not clear. Actually, you are trying to merge two different functions into one here. It might be preferable to make a new function `get_experiment_results_from_all_runs()` to make the meaning explicit.

src/quantuminspire/qiskit/qi_job.py
```diff
@@ -66,12 +77,14 @@ def submit(self) -> None:
             raise JobError('Job has already been submitted!')
         self._job_id = self._backend.run(self._qobj)
 
-    def result(self, timeout: Optional[float] = None, wait: float = 0.5) -> QIResult:
+    def result(self, timeout: Optional[float] = None, wait: float = 0.5, only_latest_run: bool = True) -> QIResult:
```
rhjdjong, line 80: You don't test for the different values of `only_latest_run`. Only the default value is tested.

rhjdjong, line 80: Similar to the comment on `get_experiment_results`, you might want to consider removing the additional argument, and in stead create a new function `all_results_from_project`, or something similarly named.

src/quantuminspire/projectq/backend_qx.py
QFer: Done

src/tests/quantuminspire/qiskit/test_quantum_inspire_provider.py
QFer: Done

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Done for all the instances

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Done

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Done

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Done

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Done

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: Done

src/quantuminspire/qiskit/qi_job.py
QFer: I thought about that but the test would actually be no different than: TestQIJob.test_result() currently is because the parameter is redirected to get_experiment_results. I would mock get_experiment_results in that test the same as done in this test, with multiple experiment_results in the return_value. The get_experiment_results is tested somewhere else for the parameter being False and True.

QFer: I can do that but this QIJob class is subclassed from a Qiskit class BaseJob where the interface is specified for getting results (job.result()) in Qiskit. I want to conform to this interface instead of extending it. Of course I can extend the interface and add a new all_results_from_project but the users won't easily find it and use it because they are known to the Qiskit interface for getting back the results. It's described this way in the Qiskit documentation etc. Only the ones who dive into it a bit and really understand how Quantum Inspire works with projects and jobs, may change (one of) the parameter of result().

src/quantuminspire/qiskit/backend_qx.py
QFer: Ok, I'll this one in get_experiment_results_from_latest_run and get_experiment_results_from_all_jobs

src/quantuminspire/qiskit/qi_job.py
rhjdjong: Hmm. I'm not sure that this argument holds. You are already extending the interface by introducing a new boolean flag as function parameter. A flag parameter to make the function behave differently is actually a well-known code smell. The recommended refactoring for that is to create a new function for the different behavior.
And in both cases (flag parameter or new function) the user must have done some research before they can use it.

QFer: I'll add another method for getting the results for all runs: results_all_runs

src/quantuminspire/qiskit/qi_job.py
```diff
@@ -77,32 +77,60 @@ def submit(self) -> None:
...
             elapsed_time = time.time() - start_time
             if timeout is not None and elapsed_time > timeout:
                 raise JobTimeoutError('Failed getting result: timeout reached.')
             time.sleep(wait)
-        if only_latest_run:
-            experiment_results = self._backend.get_experiment_results_from_latest_run(self)
-        else:
-            experiment_results = self._backend.get_experiment_results_from_all_jobs(self)
+
+    def result(self, timeout: Optional[float] = None, wait: float = 0.5) -> QIResult:
```
rhjdjong, line 97: This can be refactored further:
```
    def _result(self, result_function: Callable[[], List[ExperimentResult]], timeout: Optional[float] = None,
                wait: float = 0.5) -> QIResult:
        self._wait_for_job_finished(timeout, wait)
        experiment_results = result_function()
        return QIResult(backend_name=self._backend.backend_name, backend_version=quantum_inspire_version,
                        job_id=self.job_id(), qobj_id=self.job_id(), success=True, results=experiment_results)

    def result(self, timeout: Optional[float] = None, wait: float = 0.5) -> QIResult:
        return self._result(self._backend.get_experiment_results_from_latest_run, timeout, wait)

    def result_all_jobs(self, timeout: Optional[float] = None, wait: float = 0.5) -> QIResult:
        return self._result(self._backend.get_experiment_results_from_all_jobs, timeout, wait)


```

QFer: I don't think shorter is always better. The 'function pointers' gives additional complexity in the resulting code. But I'll change again. I undo the _wait_for_job_finished extract method refactoring.

rhjdjong (approved):

PR #94, opened 2021-04-16
Update CI files

peendebak:
* Add python 3.9 to CI
* Python 3.8 was aleady in CI, but not in `setup.py`. Added 3.8 and 3.9 to `setup.py`


peendebak:
@QFer I updated the `setup.py`. To be consistent I updated the travis configuration as well, but I am not sure it is used at all.

src/quantuminspire/api.py
```diff
@@ -813,7 +813,8 @@ def _generate_error_result(message: str) -> Dict[str, Any]:
     def execute_qasm(self, qasm: str, backend_type: Optional[Union[Dict[str, Any], int, str]] = None,
                      number_of_shots: Optional[int] = None, collect_tries: Optional[int] = None,
                      default_number_of_shots: Optional[int] = None, identifier: Optional[str] = None,
-                     full_state_projection: bool = False) -> Dict[str, Any]:
+                     full_state_projection: bool = False,
+                     user_data : str = '') -> Dict[str, Any]:
```
QFer, line 817: user_data should be added to the Args block. You can copy the text from execute_qasm_async method

QFer (changes requested):

src/quantuminspire/api.py
peendebak: The `user_data` fix perhaps should have gone into a different PR. I updated the docstring

QFer (approved):

PR #96, opened 2021-05-10
[SDS-242] Optimized-flag not valid for hardware

QFer:
* Full state projection is never True for a created job that runs on hardware
* Fixed a mypy error

### DIFF
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -501,6 +501,7 @@ class QuantumInspireAPI:
             The properties describing the new job.
             See :meth:`~.get_job` for a description of the job properties.
         """
+        full_state_projection = full_state_projection and not backend_type.get("is_hardware_backend", False)
         payload = {
             'status': 'NEW',
             'name': name,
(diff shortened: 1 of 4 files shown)


src/quantuminspire/api.py
```diff
@@ -501,6 +501,7 @@ def _create_job(self, name: str, asset: Dict[str, Any], number_of_shots: int,
             The properties describing the new job.
             See :meth:`~.get_job` for a description of the job properties.
         """
+        full_state_projection = full_state_projection and not backend_type.get("is_hardware_backend", False)
```
oschusler, line 504: Shouldn't the fallback value be `True`. If we don't know whether it is a hardware_backend or not (for whatever value), safest would be to default to see the algorithm as non-deterministic?

oschusler (changes requested):
This branch is out of date

src/quantuminspire/api.py
QFer: That's exactly why the default is False, non-deterministic (full_state_projection = False). See also knowledge base, optimizations of simulation: "Full state projection is the keyword used internally in Quantum Inspire to do optimization using a single determination of the full final state. By default, this argument is set to False"

QFer:
I'll rebase this change

src/quantuminspire/api.py
QFer: !! I see, you are right, I will change to True, I overlooked the not

oschusler (approved):

PR #99, opened 2021-05-26
[SDS-146] Split examples from sdk

QFer:
* Examples (classification and sat_example) are split from SDK and put in a separate repository (quantum-inspire-examples)
* pypi token in travis was wrong because of update
* Readthedocs added the basic examples for notebooks (projectq and qiskit)
* Some small adjustments for better documentation (Qiskit instead of QisKit)


peendebak (approved):
Voor beide repos: qi_performance_test.ipynb: de hele functie `def format_vector` wordt niet gebruikt en kan verwijderd worden. Verder akkoord met de al eerder gemaakte opmerkingen

PR #102, opened 2021-06-03
[SDS-301] Fix for damaged projects and not enough classical bits

QFer:
* Adjusted readthedocs for using version 3.8 of python
* Fix for damaged projects when error occured while post-processing the cQASM algorithm
* Check added during pre-processing for experiments where the number of classical bits is not enough to hold the measured qubits
* Removed format_vector method not used in notebook
* Renamed IPython to Jupyter

### DIFF
--- a/src/quantuminspire/qiskit/backend_qx.py
+++ b/src/quantuminspire/qiskit/backend_qx.py
@@ -284,8 +284,8 @@ class QuantumInspireBackend(BaseBackend):  # type: ignore
         number_of_classical_bits = measurements['number_of_clbits']
         max_measurement_index = max(measurement[1] for measurement in measurements['measurements'])
         if max_measurement_index >= number_of_classical_bits:
-            raise QisKitBackendError("Number of classical bits is not sufficient for storing the outcomes of the"
-                                     " experiment")
+            raise QisKitBackendError(f"Number of classical bits ({number_of_classical_bits}) is not sufficient for "
+                                     f"storing the outcomes of the experiment")
 
         if BaseBackend.configuration(self).conditional:
             number_of_qubits = experiment.header.n_qubits
@@ -391,7 +391,7 @@ class QuantumInspireBackend(BaseBackend):  # type: ignore
         return classical_state_hex
 
     @staticmethod
-    def __convert_histogram(result: Dict[str, Any], measurements: Dict[str, Any])  -> Dict[str, float]:
+    def __convert_histogram(result: Dict[str, Any], measurements: Dict[str, Any]) -> Dict[str, float]:
         """Convert histogram
 
         The quantum inspire backend always uses full state projection. The SDK user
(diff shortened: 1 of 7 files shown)


src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -275,6 +280,13 @@ def __validate_number_of_clbits(self, experiment: QasmQobjExperiment) -> None:
         if number_of_clbits < 1:
             raise QisKitBackendError("Invalid amount of classical bits ({})!".format(number_of_clbits))
 
+        measurements = self._collect_measurements(experiment)
+        number_of_classical_bits = measurements['number_of_clbits']
+        max_measurement_index = max(measurement[1] for measurement in measurements['measurements'])
+        if max_measurement_index >= number_of_classical_bits:
+            raise QisKitBackendError("Number of classical bits is not sufficient for storing the outcomes of the"
```
peendebak, line 287:
```suggestion
            raise QisKitBackendError(f"Number of classical bits {number_of_classical_bits} is not sufficient for storing the outcomes of the"
```

src/tests/quantuminspire/qiskit/test_backend_qx.py
```diff
@@ -260,6 +260,32 @@ def test_run_returns_correct_result_for_my_project_number(self):
...
+        default_project_number = 42
+        my_project_number = 43
+        api = Mock()
+        api.create_project.return_value = {'id': default_project_number}
+        quantum_inspire_job = Mock()
+        quantum_inspire_job.get_project_identifier.return_value = my_project_number
+        api.execute_qasm_async.return_value = quantum_inspire_job
+        api.get_backend_type_by_name.return_value = {'max_number_of_shots': 4096}
+        simulator = QuantumInspireBackend(api, Mock())
+        instructions = [{'name': 'h', 'qubits': [0]},
```
peendebak, line 273: Are all these instructions needed? This is internal qiskit format and prone to change. Could you not write this test by letting qiskit generarate similar data? E.g.
```
c=QuantumCircuit(3,3)
c.h(0)
c.h(2)
c.measure_all()
...
```

peendebak (commented):
Two minor comments, otherwise looks ok

src/quantuminspire/qiskit/backend_qx.py
QFer: Changed

src/tests/quantuminspire/qiskit/test_backend_qx.py
QFer: All these instructions are not necessary, I copied them from another test. Actually a measurement is enough, I changed that.
You are right that this is qiskit format but the easiest way to create sometimes specific instructions for the various test cases. For some test cases this is actually not needed, I added a ticket to the SDK backlog to investigate this suggestion.

peendebak (approved):

PR #103, opened 2021-06-03
[SDS-307] Now able to get calibration data from result

QFer:
* Added method to get calibration info to api
* Added unit tests

### DIFF
--- a/src/quantuminspire/api.py
+++ b/src/quantuminspire/api.py
@@ -49,7 +49,7 @@ class QuantumInspireAPI:
                  coreapi_client_class: Type[coreapi.Client] = coreapi.Client) -> None:
         """ Python interface to the Quantum Inspire API (Application Programmer Interface).
 
-        The Quantum Inspire API supplies an interface for executing cQASM programs and can be used to access the
+        The Quantum Inspire API supplies an interface for executing programs and can be used to access the
         different entities in the Quantum Inspire database needed for running the programs.
         The entities for which an interface is provided are:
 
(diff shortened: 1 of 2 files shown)


src/quantuminspire/api.py
```diff
@@ -676,6 +676,32 @@ def get_measurement_register_from_result(self, result_id: int) -> List[Any]:
             raise ApiError(f'Measurement register for result with id {result_id} does not exist!') from err_msg
         return measurement_register
 
+    def get_calibration_from_result(self, result_id: int) -> Optional[Any]:
+        """ Gets the calibration information of the executed cQASM code, given the result_id.
```
peendebak, line 680:
```suggestion
        """ Gets the calibration information of the executed code, given the result_id.
```
We also submit qiskit code (right that that is converted to cqasm internally, but that might change)

src/quantuminspire/api.py
```diff
@@ -676,6 +676,32 @@ def get_measurement_register_from_result(self, result_id: int) -> List[Any]:
...
+
+        :return:
+            The calibration information from the backend.
+            None is returned when there is no calibration information (e.g. simulator backend).
+        """
+        calibration_info = None
+        result = self.get_result(result_id)
+        calibration_url = result.get('calibration')
+        if calibration_url is not None:
+            try:
```
peendebak, line 695: I guess this code works, but is the parsing of the url needed? One can just do
```
result = requests.get(c, auth=authentication)
calibration = result.json()
```
(the `authentication` and `.json` can be wrapped in something off course)

QFer: Yes, you can directly use request package to send a get for this url,, but I think it is better to use the already authenticated coreapi consistently for all communication with the backend api as it is now done in SDK.

src/quantuminspire/api.py
QFer: Changed it to executed job. Also changed some other instances of cQASM (where it is not needed to be called cQASM explicitly)

peendebak (dismissed):
Functionality here is good. There is still the issue of documenting the (possible) confusion between the job_id and project_id, but that should be done in another PR.

QFer:
Had to force push after rebase

oschusler (approved):

Issue #107, opened 2021-08-02
Qiskit SDK fails for Starmon-5


peendebak:
# Bug report

## What did you try to do?

Run the following example:

```
from quantuminspire.credentials import get_authentication
from quantuminspire.qiskit import QI
from qiskit import execute, Aer
from qiskit.circuit import QuantumCircuit
from qiskit.utils import QuantumInstance

authentication = get_authentication()
QI.set_authentication(authentication)
api = QI.get_api()
backend = QI.get_backend("Starmon-5")
#backend=Aer.get_backend('aer_simulator')

quantum_instance=QuantumInstance(backend=backend, shots=200)

qc=QuantumCircuit(2)
qc.h(0)
qc.cz(0,1)
qc.barrier()
qc.measure_all()

result=quantum_instance.execute([qc])
```

## What did you expect to happen?

The code should have been executed without an error. 

## What actually happened?

An error was generated:
```
...
  File "c:\projects\qiskit-terra\qiskit\transpiler\preset_passmanagers\level1.py", line 231, in level_1_pass_manager
    if coupling_map and not coupling_map.is_symmetric:
AttributeError: 'tuple' object has no attribute 'is_symmetric'
```

The problem seems to be in the QI backends not generating a proper `CouplingMap`, but just a plain python list of tuples.
Fixing that results in another error:
```
AttributeError: 'QIJob' object has no attribute 'queue_position'
```

## Environment

Please specify:
* operating system, version and distribution: Windows 10
* python version (`python --version`) and distribution: Python 3.8

@oschusler @QFer

QFer:
@peendebak two problems with running the latest version of qiskit terra.
1. Coupling map was list of tuples but instead it should be a list of lists.
2. queue_position method is called when we return status of the job in the backend is QUEUED (which is a valid state). We should provide it in our Job class.

QFer:
Solved in https://github.com/QuTech-Delft/quantuminspire/pull/109

PR #109, opened 2021-08-23
[SDS-423] Coupling map qiskit

QFer:
* Fixed coupling map type
* Added queue_position method to job for queued jobs
* Calculate total time taken for all jobs in result


codecrap (approved):
I reviewed the code and ran it against the minimal example in issue #107, and I do not get the error anymore so the issue seems solved. @peendebak

eendebakpt (approved):

Issue #113, opened 2021-08-31
QI Qiskit SDK returns incorrect status object


eendebakpt:
# Bug report

## What did you try to do?

Run the following minimal example:
![image](https://user-images.githubusercontent.com/883786/131476385-d51923e7-d023-4966-a1ea-5d3314908bfe.png)

The result is not correct:

1. The `status` field is not set, while at the time of execution the `Starmon-5` backend was offline
2. The field `pending jobs` is incorrect. It reports zero, but there is at least 1 pending job:
![image](https://user-images.githubusercontent.com/883786/131476584-d8f57ff4-4236-456d-8f49-4685be91579d.png)

## What did you expect to happen?

The status object should either return correct results, or fill the values with a dummy. (e.g. for `pending jobs`  a `NaN`, or `None` or `-1`.

@QFer

QFer:
Agree, this BackendStatus structure we should provide. The pending_jobs is currently not known so we will insert -1 as it is defined as an int

QFer:
Solved in https://github.com/QuTech-Delft/quantuminspire/pull/114

PR #114, opened 2021-08-31
[SDS-NNN] Backend status

QFer:
* Added status() to backend to return the BackendStatus
* pending jobs are not known. We leave it 0 because Qiskit doesn't accept negative values.
* backend_version is not coming from the backend, currently SDK version is taken as backend version


QFer:
Solves https://github.com/QuTech-Delft/quantuminspire/issues/113

peendebak (dismissed):

eendebakpt (approved):

PR #117, opened 2021-10-04
[SDS-371] Qiskit circuits in tests

QFer:
* In the circuit parser tests and the Qiskit backend tests we now build the circuits that are used for generating input instead of the internals of Qiskit which may change in a future version.
* u1, u2 and u3 are deprecated in Qiskit since 0.16.0 but still supported until these gates are really deleted from Qiskit.


abousias (approved):
Only one small less important comment. The test assertions have  stayed the same and it looks good. In the future we can look into separating the tests into even smaller units, like for example testing `CircuitToString` without the involvement of `QuantumInspireBackend` and maybe try to shorten the tests (if possible).

PR #120, opened 2021-11-08
New CalibrationViewer helper object for nice printing of backend calibration data

codecrap:
This is a proposal for a CalibrationViewer helper object to nicely handle and print backend calibration data.
It uses rich printing but falls back to the builtin print if not available.
I will add more tests and further refine the class if it is in principle okay to include this helper object in the main repo.

An example of its use is currently available in [the examples repo](https://github.com/QuTech-Delft/quantum-inspire-examples/blob/feat/example_calibration_retrieval/docs/notebooks/example_calibration_data.ipynb) ([updated version here](https://github.com/codecrap/quantum-inspire-examples/blob/feat/example_calibration_retrieval/docs/notebooks/example_calibration_data.ipynb)).


codecrap:
The pipeline is failing because it tries to run a notebook that requires authentication, which it doesn't receive and thus the build times out. The same test does not seem to be a problem for #121 (https://github.com/QuTech-Delft/quantuminspire/pull/121). Can you help out with this issue @QFer ?

QFer:
@codecrap The reason is that pull requests sent from forked repositories do not have access to encrypted authentication secrets needed for running the tests on quantum inspire backend.
A possible solution is to disable these example runs for pull requests.

QFer:
@codecrap I made a change so the pull request coming from a fork is not running the examples.  While reviewing and before merging we will run the examples and check correctness. Please add unit test to coverage 100% and I guess your (out-of-date) branch needs a rebase or merge.

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,124 @@
...
+                        rprint(f"\t{key}: {item['value']} [{item['unit']}]")
+                    else:
+                        rprint(f"\t{key}: {item}")
+                else:
+                    rprint(f"Unknown structure '{key}' in qubit calibration.")
+
+    def get_calibration_field(self, field: str) -> Any:
+        """
+        Safely gets field `field` from system calibration. 
+
```
peendebak, line 82: Add docstring for parameter `field`

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,124 @@
...
+        if field not in self.calibration['parameters']['system']:
+            raise KeyError(f"Calibration field '{field}' does not exist in system calibration dictionary.")
+
+        return self.calibration['parameters']['system'][field]
+
+    def show_calibration_field(self, field: str) -> Any:
+        """
+        Print calibration for a specific key `field` and return the content.
+        Also prints standard subfields 'timestamp' and 'status', if they are available.
+
```
peendebak, line 96: Missing docstring for `field`

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        self._calibration = calibration_dictionary
+        return
+
+    @property
+    def calibration(self) -> Dict[str, Any]:
+        return self._calibration
+
+    @property
+    def timestamp(self) -> str:
```
peendebak, line 44: Add docstring

src/quantuminspire/calibration_viewer.py
codecrap: Fixed in new commit

src/quantuminspire/calibration_viewer.py
codecrap: Fixed in new commit

src/quantuminspire/calibration_viewer.py
codecrap: Fixed in new commit

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+
+        for key, item in system.items():
+            if isinstance(item, dict):
+                if np.all([k in item for k in ['unit', 'value']]):
+                    rprint(f"\t{key}: {item['value']} [{item['unit']}]")
+                else:
+                    try:
+                        rprint(f"\t{key}: calibration at {item['timestamp']}")
+                    except KeyError:
+                        rprint(f"Key 'timestamp' not in item {item}.")
```
QFer, line 62: 1. the method is called: show_system_parameters so I expect the system structure to be printed. But what I get is only a timestamp?
2. when timestamp is not in the structure we get lines with key 'timestamp' not in messages. Why not just print item in this case as I would expect from the method name (and as done in show_qubit_parameters)

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+            else:
+                rprint(f"Unknown structure '{key}' in system calibration.")
```
QFer, line 64: Actually key is not the unknown structure but item is. The message is misleading.

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+        for qubit, data in qdata.items():
+            rprint(f'[green]Qubit[/green]: {qubit}')
+            for key, item in data.items():
+                if isinstance(item, dict):
+                    if np.all([k in item for k in ['unit', 'value']]):
+                        rprint(f"\t{key}: {item['value']} [{item['unit']}]")
+                    else:
+                        rprint(f"\t{key}: {item}")
+                else:
+                    rprint(f"Unknown structure '{key}' in qubit calibration.")
```
QFer, line 79: Same here: key is not the unknown structure but item is. The message is misleading.

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Print calibration for a specific key `field` and return the content.
+        Also prints standard subfields 'timestamp' and 'status' if they are available.
+
+        :param field: The entry (dict key) of the system calibration dictionary whose value should be printed.
+
+        :return: The system calibration content of `field`.
+        """
+        simple_keys = ['timestamp', 'status']
+        cal = self.get_calibration_field(field)
```
QFer, line 106: cal -> calibration_field

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+            try:
+                rprint(f"\t[green]{key}[/green]: {cal[key]}")
+            except (KeyError, TypeError):
+                pass
+
+        other_fields = set(cal.keys()) - set(simple_keys)
+        rprint(f"[green]other fields[/green]:", end='')
+        rprint(f"\t{str(other_fields)}")
+
+        return cal
```
QFer, line 123: This return is odd. For this we have method get_calibration_field. I don't expect a method that is called show_calibration_field to return something.

src/tests/quantuminspire/test_calibrationviewer.py
```diff
@@ -0,0 +1,158 @@
...
+                                                    {'value': '20.9',
+                                                     'unit': 'mK',
+                                                     'symbol': 'Fridge temperature'},
+                                               'last_calibration_date':
+                                                   {'value': '04/11/2021 - 16:42 PM',
+                                                    'unit': 'iso-8601',
+                                                    'symbol': 'Calibration date'},
+                                                'extra_field':
+                                                    {'number': 0},
+                                                'unknown_field': [0,1,2]
```
QFer, line 42:
```suggestion
                                                'unknown_field': [0, 1, 2]
```

src/tests/quantuminspire/test_calibrationviewer.py
```diff
@@ -0,0 +1,158 @@
...
+                                                        {'value': 99.7,
+                                                         'unit': '%',
+                                                         'symbol': 'Finit'},
+                                                    'readout_fidelity':
+                                                        {'value': 94.8,
+                                                         'unit': '%',
+                                                         'symbol': 'FR/O'},
+                                                    'extra_field':
+                                                        {'number': 0},
+                                                    'unknown_field': [0,1,2]
```
QFer, line 73:
```suggestion
                                                    'unknown_field': [0, 1, 2]
```

src/tests/quantuminspire/test_calibrationviewer.py
```diff
@@ -0,0 +1,158 @@
...
+        output = s.getvalue()
+        self.assertIn('two_qubit_gate_fidelity', output)
+        self.assertIn('98.1', output)
+        self.assertIn('t1', output)
+        self.assertIn('t2star', output)
+        self.assertIn('extra_field', output)
+        self.assertIn('0', output)
+        self.assertIn('Unknown structure', output)
+
+
```
QFer, line 114: remove 1 newline

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+
+        for key in simple_keys:
+            try:
+                rprint(f"\t[green]{key}[/green]: {cal[key]}")
+            except (KeyError, TypeError):
+                pass
+
+        other_fields = set(cal.keys()) - set(simple_keys)
+        rprint(f"[green]other fields[/green]:", end='')
+        rprint(f"\t{str(other_fields)}")
```
QFer, line 121: When I call a method show_calibration_field I expect that the fields are printed but now it only prints some predefined values and the other key values of the structure and not the values? What is the idea behind this?

Why not just print the calibration field as expected?

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+        :return: The system calibration content of `field`.
+        """
+        if field not in self.calibration['parameters']['system']:
+            raise KeyError(f"Calibration field '{field}' does not exist in system calibration dictionary.")
+
+        return self.calibration['parameters']['system'][field]
+
+    def show_calibration_field(self, field: str) -> Any:
+        """
+        Print calibration for a specific key `field` and return the content.
```
QFer, line 98: This is what I expect but it doesn't do this. It doesn't print the field when 'value' key is part of the structure.

src/quantuminspire/calibration_viewer.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        simple_keys = ['timestamp', 'status']
+        cal = self.get_calibration_field(field)
+        rprint(f'[green]Calibration[/green]: {field}')
+        try:
+            rprint(f"\t{cal['value']}")
+        except (KeyError, TypeError):
+            rprint(f"\t{cal}.")
+
+        for key in simple_keys:
```
QFer, line 113: When 'value' is not part of cal the complete structure is printed. Why bother printing these values again

src/quantuminspire/calibration_viewer.py
QFer, line 106: This call to get_calibration_field may raise an exception. Do you want this show_calibration_field raise this exception when a non-exiting field is entered or handle it differnetly?

QFer (changes requested):

PR #127, opened 2021-12-17
[SDS-NNN] Changes needed to display correct html again

QFer:
* Changes in conf.py and css file in _static\theme_overrides to:
  * display correct html output
  * correct for the wide tables

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ To install the necessary packages to perform documentation activities for SDK do
 pip install -e .[rtd]
 ```
 
+The documentation generation process is dependent on pandoc. When you want to generate the
+documentation and pandoc is not yet installed on your system navigate
+to [Pandoc](https://pandoc.org/installing.html) and follow the instructions found there to install pandoc.
 To build the 'readthedocs' documentation do:
 
 ```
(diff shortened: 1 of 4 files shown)


rhjdjong:
Building the documentation on windows fails, because pandoc is not installed by default:
```
> pip install -e .[rtd]
> cd docs
> make html
Running Sphinx v4.3.2
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 20 source files that are out of date
updating environment: [new config] 20 added, 0 changed, 0 removed
reading sources... [ 25%] examples/example_projectq                            
Notebook error:
PandocMissing in examples\example_projectq.ipynb:
Pandoc wasn't found.
Please check that pandoc is installed:
https://pandoc.org/installing.html
```

`pip install pandoc` only installs the python wrapper functions for the pandoc executable. The actual pandoc executable must have been installed already in the system. This should be mentioned in the README.
After installing pandoc via the website mentioned in the message, the documentation build succeeds.

rhjdjong (changes requested):
See general comment about the need to have pandoc installed for documentation builds

Issue #145, opened 2022-07-15
Calculating Quantum Volumes


QFer:
We want to calculate metrics on different backends, under which the QI backends. We created an implementation based upon qiskit-experiments package to calculate quantum volume. Running this on QI backends gives errors.

![image](https://user-images.githubusercontent.com/46991941/179211007-8ca2d3cc-b0c5-4ec9-885c-deef83304d39.png)

This minimal example program should run (as it does on other similar backend from IBM)
```
# *** Necessary modules
from qiskit_experiments.framework import BatchExperiment
from qiskit_experiments.library import QuantumVolume

from quantuminspire.qiskit import QI
QI.set_authentication()

# *** Choose actual backend
backend_id = QI.get_backend('QX single-node simulator')

# *** Choose min and max qubits
min_qubits = 2
max_qubits = 4

# *** Compute Quantum Volume
exps = [QuantumVolume(range(1,i+1), trials=100) for i in range(min_qubits, max_qubits+1)]
batch_exp = BatchExperiment(exps)
batch_exp.set_transpile_options(optimization_level=3)

# Run
batch_expdata = batch_exp.run(backend_id).block_for_results()

qv_values = [
    batch_expdata.child_data(i).analysis_results("quantum_volume").value
    for i in range(batch_exp.num_experiments)
]

print(f"Quantum volume is: {max(qv_values)}")
```

**Acceptance criteria**
 - if this feature is available the code example above runs correctly

QFer:
Solution is merged to dev-branch and will be part of the next release

PR #146, opened 2022-07-15
Draft: [SDS-716] feature quantum volume

QFer:
* Solves issue https://github.com/QuTech-Delft/quantuminspire/issues/145
* Added allow_fsp as run_config parameter. Default True, when allow_fsp = False then no full_state_projection for running qiskit algorithm.
* TODO: unit test
* TODO: design decision to be taken on serializer positioning


QFer:
#147 replaces this pull request

PR #147, opened 2022-09-16
[SDS-716] Quantum Volume Computation

QFer:
* Solves issue https://github.com/QuTech-Delft/quantuminspire/issues/145
* user data added more header data for usage in result handling
* metadata from qiskit header needs to be serialized. Because of numpy arrays we use serializer from qilib
* qilib added to setup
* Some small pylint improvements, f-strings and setting encoding for opening files
* Added parameter allow_fsp to run in order to prevent fsp (manually) for experiments


src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -173,17 +178,19 @@ def run(self,
         self.__validate_number_of_shots(number_of_shots)
 
         identifier = uuid.uuid1()
-        project_name = 'qi-sdk-project-{}'.format(identifier)
+        project_name = f'qi-sdk-project-{identifier}'
         project: Optional[Dict[str, Any]]
         project = self.__api.create_project(project_name, number_of_shots, self.__backend)
+        if not allow_fsp:  # when allow_fsp = False, we know what we do, no message needed
```
tomrijnbeek, line 184: I don't really understand what this comment means. "we know what we do" is very vague to me, and probably requires understanding what `show_fsp_warning` does exactly. Clearly if FSP is not allowed there is something to be aware of (maybe if you never specify a measure operation or something?). It would be good to clarify this comment.

src/quantuminspire/qiskit/backend_qx.py
```diff
@@ -255,18 +261,35 @@ def _generate_cqasm(self, experiment: QasmQobjExperiment, measurements: Measurem
...
+
+        :return:
+            A structure with user data that is needed to process the result of the experiment.
+        """
+        user_data = {'name': experiment.header.name, 'metadata': serializer.serialize(experiment.header.metadata),
+                     'qubit_labels': experiment.header.qubit_labels, 'qreg_sizes': experiment.header.qreg_sizes,
+                     'clbit_labels': experiment.header.clbit_labels, 'creg_sizes': experiment.header.creg_sizes,
+                     'global_phase': experiment.header.global_phase, 'memory_slots': experiment.header.memory_slots,
+                     'measurements': measurements.to_dict()}
+        return user_data
```
tomrijnbeek, line 285: nit: what is the reason we don't just inline the return? The assignment of the local variable is never used besides returning it immediately.

src/quantuminspire/version.py
```diff
@@ -1 +1 @@
-__version__ = '2.0.0'
+__version__ = '2.1.0'
```
tomrijnbeek, line 1: Some of the changes in this PR appear to be breaking changes. Specifically, I notice that the default value of `full_state_projection` was switched in some places, and the dependency changes are also breaking since we dropped support for Python 3.7. By SemVer 2.0, shouldn't the major version be increased?

tomrijnbeek (changes requested):

src/quantuminspire/qiskit/backend_qx.py
QFer: This flag is set to true only when the user added explicitly "allow_fsp=False" as a parameter to the Qiskit execute method to run the circuit. These non-qiskit parameters are stored in the run_config. In method run (in backend_qx) we handle allow_fsp when it was added to run_config.
In some cases (for example to compute quantum volume) the user never wants to run full state projection (an automatic optimization in our simulation backends) even when the circuit can be optimized. For that I added this flag. When it is false the user did that (that's why I added we know what we do....).
I added the following comment instead:
method run was called explicitly with allow_fsp = False. That is: full_state_projection is turned off.
No need to attend the user that the execution may take longer.

src/quantuminspire/qiskit/backend_qx.py
QFer: None actually. I will change it

src/quantuminspire/version.py
QFer: We didn't drop support for python 3.7 (see setup.py). We only dropped the travis build for python 3.7 (because dependencies problems on their environment).

tomrijnbeek (approved):

PR #432, opened 2026-04-16
[QI2-1729] connect cli and api

nienkeduetz:
[Note: The polling of job logs currently gives an error on the local environment, as we do not yet have a local version of Loki.](https://github.com/QuTech-Delft/quantuminspire/pull/431)


quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+@files_app.command("compile")
+def compile_file(
+    file: Optional[Path] = typer.Option(None, help="The path to the algorithm file to compile."),
+    name: Optional[str] = typer.Option(
+        None, help="The name of a previously initialized algorithm whose settings will be used."
+    ),
+    backend_type_id: Optional[int] = typer.Option(None, help="The ID of the backend type to target."),
+    compile_stage: Optional[CompileStage] = typer.Option(
+        None, help="The stage up to which the algorithm should be compiled."
+    ),
```
hjir, line 123: I think it would be useful to add which one is default? based on my tries it's the decomposition one?

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+) -> None:
+    """Compile an algorithm file."""
```
hjir: Maybe we can already mention here that this only works for quantum algorithms, not hybrid ones?

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
 def run_job(
-    file: str = typer.Argument(..., help="The path to the hybrid algorithm or quantum circuit."),
-    name: Optional[str] = typer.Option(None, help="The name of the file."),
-    num_shots: Optional[str] = typer.Option(None, help="The Number of shots"),
-    store_raw_data: Optional[str] = typer.Option(None, help="Whether to store data for each shot."),
-    backend_type_id: Optional[int] = typer.Option(None, help="The backendtype id."),
+    file: Optional[Path] = typer.Option(None, help="The Path to the hybrid algorithm or quantum circuit."),
+    backend_type_id: Optional[int] = typer.Option(None, help="The backend type id."),
+    num_shots: Optional[int] = typer.Option(None, help="The Number of shots"),
+    store_raw_data: Optional[bool] = typer.Option(None, help="Whether to store data for each shot."),
```
hjir: Include what is the default option? This shows the two flags in the documentation (--store-raw-data and --no-store-raw-data) but it's not clear which one is default

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+) -> None:
+    """Retrieve information about a specific backend type."""
+    api = Api()
+    typer.secho(f"Retrieving backend type with ID '{backend_type_id}'...", fg=typer.colors.BLUE)
+    backend_type = api.get_backend_type(backend_type_id)
+    console.print(backend_type.model_dump())
 
 
 @jobs_app.command("run")
 def run_job(
```
hjir, line 173: Generally here I had trouble figuring out from the `--help` documentation which arguments are optional and which not. It says `Usage: qi jobs run [OPTIONS]` but doesn't say what happens if no options are given?

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+    algorithm_name: Optional[str] = typer.Option(
+        None, help="The name of the algorithm. The results of the latest job for the algorithm will be retrieved."
+    ),
+    save: Optional[bool] = typer.Option(None, help="Save the results to a file."),
 ) -> None:
     """Retrieve information about a specific result."""
-    typer.secho(f"Retrieving result with ID '{id}'...", fg=typer.colors.BLUE)
-    typer.echo("Result details: {...}")  # Placeholder for actual result details
+    api = Api()
+    typer.secho(f"Retrieving result with ID '{job_id}'...", fg=typer.colors.BLUE)
```
hjir: This can be a bit confusing when you put in `algorithm_name` but no `job_id` because then it says None

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+    api = Api()
+    status = api.get_job_status(algorithm_name, job_id, wait, timeout)
+    if job_id is not None:
+        typer.echo(f"Job {job_id} status: '{status.value}'")
+    if algorithm_name is not None:
+        typer.echo(f"Latest job of {algorithm_name} status: '{status.value}'")
+
+
+@jobs_app.command("results")
+def get_results(
```
hjir, line 246: I had an issue with this where I got results if I put in the job_id but didn't get anything when I put in the algorithm_name (and I did define it at `qi jobs run`) - does it need to be persisted to be able to use the algorithm name?

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
 @jobs_app.command("final-result")
 def get_final_result(
-    id: Optional[int] = typer.Argument(None, help="The ID of the job to retrieve the result for."),
+    job_id: Optional[int] = typer.Option(None, help="The ID of the job to retrieve the result for."),
+    algorithm_name: Optional[str] = typer.Option(
+        None, help="The name of the algorithm. The final result of the latest job for the algorithm will be retrieved."
+    ),
+    save: Optional[bool] = typer.Option(None, help="Save the final result to a file."),
 ) -> None:
     """Retrieve information about a specific final result."""
```
hjir: I know this is not actually part of this PR, but maybe it'd be nice to say what is the difference between results and final-result?

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
     """
-    typer.echo("Logout successful!")
+    api = Api()
+    api.logout(hostname)
+
+
+@projects_app.command("init")
+def initialize_project(
+    name: str = typer.Argument(None, help="The project name"),
+    description: str = typer.Argument("", help="The project description"),
```
hjir: I'm not sure why but the usage for this command kinda looks like name and description are optional:

`Usage: qi projects init [OPTIONS] [NAME] [DESCRIPTION]`

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+    table.add_row("Source", job.source or "N/A")
 
+    console.print(table)
 
-@jobs_app.command("result")
-def get_result(
-    id: Optional[int] = typer.Argument(None, help="The ID of the job to retrieve the result for."),
+
+@jobs_app.command("status")
+def get_job_status(
```
hjir, line 224: Here both `job_id` and `algorithm_name` are set to optional but we need to define at least one?

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+    wait: bool = typer.Option(False, help="Wait for the job to complete."),
+    timeout: float = typer.Option(60, help="Timeout for the job to complete."),
+) -> None:
+    """Retrieve the status of a job."""
+    api = Api()
+    status = api.get_job_status(algorithm_name, job_id, wait, timeout)
+    if job_id is not None:
+        typer.echo(f"Job {job_id} status: '{status.value}'")
+    if algorithm_name is not None:
+        typer.echo(f"Latest job of {algorithm_name} status: '{status.value}'")
```
hjir:
```suggestion
        typer.echo(f"Status of latest job for {algorithm_name} algorithm: '{status.value}'")
```
Would something like maybe be clearer? I had an algorithm called "hybrid" and "latest job of hybrid status" was a bit confusing 😇

quantuminspire/cli.py
hjir: From what I tested, it does need to be with the --persist flag to later use the algorithm-name as an identifier. Could you make this clear in the docs for the flag (something like "This will allow you to get results based on the name of the algorithm.")?

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
+        None, help="The name of the algorithm. The final result of the latest job for the algorithm will be retrieved."
+    ),
+    save: Optional[bool] = typer.Option(None, help="Save the final result to a file."),
 ) -> None:
     """Retrieve information about a specific final result."""
-    typer.secho(f"Retrieving final result with ID '{id}'...", fg=typer.colors.BLUE)
-    typer.echo("Final result details: {...}")  # Placeholder for actual final result details
+    api = Api()
+
+    typer.secho(f"Retrieving final result with ID '{job_id}'...", fg=typer.colors.BLUE)
```
hjir: Same as the above comment for results, if job_id is not defined but algorithm_name is, this log is confusing because it says None

quantuminspire/cli.py
```diff
@@ -32,72 +49,310 @@ def logout(
...
     """Run a new job on Quantum Inspire."""
-    typer.echo(f"Running job for file '{file}'...")
+    api = Api()
+    api.execute_algorithm(file, backend_type_id, num_shots, store_raw_data, name, persist)
+
+    if file is not None:
+        typer.echo(f"Running job for file '{file}'...")
     if persist:
         typer.echo("The project and algorithm have been stored.")
     typer.echo("Job submitted successfully!")
```
hjir: Would it make sense to here return the job id as well? I don't think the user can get it any other way?

hjir (changes requested):
I think it looks good! Tested out all the commands and have some remarks on where documentation could be clearer (I am coming into this ~~clueless~~ with fresh eyes) but it shouldn't require any big changes so don't get scared by the number of comments ;)

quantuminspire/cli.py
nienkeduetz: Description can be optional, but name cannot. I have removed the default value so the user gets a Missing argument 'NAME' error.

quantuminspire/cli.py
nienkeduetz: ✅

quantuminspire/cli.py
nienkeduetz: Good idea!

quantuminspire/cli.py
nienkeduetz: Yeah this is a but confusing, I agree. The user needs to add either a file to run, or a locally persisted algorithm (which has a file in its parameters). I have added this to the doc string.

quantuminspire/cli.py
nienkeduetz: ✅

quantuminspire/cli.py
nienkeduetz: I removed all the type.echos here, because the execute_algorithm already does that where necessary. It prints the job id when the persist flag is False, so no need to print it here again.

quantuminspire/cli.py
nienkeduetz: Yes, the check for this is done in get_job_status of the Api. I have added this requirement to the doc string so it is clear to the user as well.

quantuminspire/cli.py
nienkeduetz: ✅

quantuminspire/cli.py
nienkeduetz: Agreed, I have added a case for when the user uses the algorithm name

quantuminspire/cli.py
nienkeduetz: ✅

quantuminspire/cli.py
nienkeduetz: ✅

quantuminspire/cli.py
nienkeduetz: That's correct, the algorithm_name is for the locally persisted algorithm. qi j

quantuminspire/cli.py
hjir: I can't find the change for this--If the default for store_raw_data is false, maybe it can be defined in the same way as persist?

quantuminspire/cli.py
hjir: I know that with persist, you can use the algorithm name, but I would still find it useful to print out the id? I don't have that strong of an opinion about this though, so if you disagree for any reason, all good!

quantuminspire/cli.py
nienkeduetz: Good catch, I only changed it in the initialize_algorithm() function. I have changed it here as well now.

quantuminspire/cli.py
nienkeduetz: I agree! I have changed it in api.py in the execute_algorithm() function

hjir: 🙌

hjir (approved):
Good job, approved!

PR #434, opened 2026-04-20
[QI2-1819] Conda installation mechanism

hjir:
Introduces an `conda-environment.yml` file that simplifies installation of `quantuminspire` in Conda.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ conda env create -f https://github.com/QuTech-Delft/quantuminspire/blob/feature/
 By default, the environment that will be created will be named `quantuminspire`; use the `-n/--name <CUSTOM-ENV-NAME>` option to overwrite this.
 
 > [!NOTE]  
-> Currently, both `qiskit-quantuminspire` and `pennylane-quantuminspire` are installed into your virtual environment. If this is something you don't want, you can download the `conda-environment.yml` file and manually edit the dependencies to remove them. Then you can create your environment using the command above, with the actual path to the file instead of the URL.
+> Currently, both `qiskit-quantuminspire` and `pennylane-quantuminspire` are installed into your virtual environment. If this is something you don't want, you can either `pip uninstall` these packages or download the `conda-environment.yml` file and manually edit the dependencies. Then you can create your environment using the command above, with the actual path to the edited file instead of the URL.
 
 After creating the environment, you need to activate it (use your custom environment name if you have defined it):
 
--- a/README.md
+++ b/README.md
@@ -29,13 +29,13 @@ Check you Conda version with `conda --version`, then run the appropriate command
 **Conda >= 26.3:**
 
 ```bash
-conda create --file https://github.com/QuTech-Delft/quantuminspire/blob/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
+conda create --file https://raw.githubusercontent.com/QuTech-Delft/quantuminspire/refs/heads/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
 ```
 
 **Conda < 26.3:**
 
 ```bash
-conda env create -f https://github.com/QuTech-Delft/quantuminspire/blob/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
+conda env create --file https://raw.githubusercontent.com/QuTech-Delft/quantuminspire/refs/heads/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
 ```
 
 By default, the environment that will be created will be named `quantuminspire`; use the `-n/--name <CUSTOM-ENV-NAME>` option to overwrite this.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,22 +7,48 @@ Currently, functionality of the tool is still limited, but the tool is required
...
 ```
 
 Afterwards, running `qi --help` should show a help menu.
 
-> [!WARNING]  
-> Installing quantuminspire using pipx **inside a conda environment** may cause conflicts with other packages. This can result in unexpected behavior and prevent you from using the tool correctly. To avoid these issues,
-> it is strongly recommended to install and run pipx on a system Python installation instead, downloaded from the official Python website.
-
 > [!NOTE]  
 > If your python version is newer then supported by quantuminspire, please run the following command: `pipx install quantuminspire --python /path/to/python3.12`.
```
hjir, line 19: It seems like `quantuminspire` should now support python>=3.10,<3.14, is the python3.12 in this command arbitrary?

README.md
```diff
@@ -41,7 +41,7 @@ conda env create -f https://github.com/QuTech-Delft/quantuminspire/blob/feature/
 By default, the environment that will be created will be named `quantuminspire`; use the `-n/--name <CUSTOM-ENV-NAME>` option to overwrite this.
 
 > [!NOTE]  
-> Currently, both `qiskit-quantuminspire` and `pennylane-quantuminspire` are installed into your virtual environment. If this is something you don't want, you can download the `conda-environment.yml` file and manually edit the dependencies to remove them. Then you can create your environment using the command above, with the actual path to the file instead of the URL.
+> Currently, both `qiskit-quantuminspire` and `pennylane-quantuminspire` are installed into your virtual environment. If this is something you don't want, you can either `pip uninstall` these packages or download the `conda-environment.yml` file and manually edit the dependencies. Then you can create your environment using the command above, with the actual path to the edited file instead of the URL.
```
hjir, line 44: A bit of context here: If all three packages are installed from within the .yml file, dependency conflicts will be automatically resolved. On the other hand, if we e.g. only have `quantuminspire` in the .yml file, and then from the conda environment try running `pip install qiskit-quantuminspire`, dependendency confilcts will pop up. I figure this is easier than including instructions on how to add the `qiskit-quantuminspire` and `pennylane-quantuminspire` packages to the .yml file.

README.md
```diff
@@ -29,13 +29,13 @@ Check you Conda version with `conda --version`, then run the appropriate command
...
 ```bash
-conda create --file https://github.com/QuTech-Delft/quantuminspire/blob/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
+conda create --file https://raw.githubusercontent.com/QuTech-Delft/quantuminspire/refs/heads/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
 ```
 
 **Conda < 26.3:**
 
 ```bash
-conda env create -f https://github.com/QuTech-Delft/quantuminspire/blob/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
+conda env create --file https://raw.githubusercontent.com/QuTech-Delft/quantuminspire/refs/heads/feature/QI2-1819/conda-installation-mechanism/conda-environment.yml
```
hjir, line 38: Once merged, the URL should be:

https://raw.githubusercontent.com/QuTech-Delft/quantuminspire/refs/heads/master/conda-environment.yml

Maybe this can be simplified with SiteDetour?

README.md
oschusler: yes, it is

README.md
oschusler: This is now reachable as https://conda.quantum-inspire.com/cli (I figured we would have environment files for qiskit and pennylane as well).

The apex (/) however is difficult. Currently pointing to www.quantum-inspire.com, but should maybe be something different.

README.md
oschusler: I actually don't think anyone will ever install AND qiskit AND pennylane at the same time. Also, they might want to install those in their own environments. From that perspective, it might be better to leave them to the end-user?

hjir: Okay, fair enough. Would it make sense to only install quantuminspire with the environment.yml file provided in this repository, and maybe have qiskit-quantuminspire and pennylane-quantuminspire .yml files in their respective repositories? 
(So then I could also adjust the installation instructions in those repositories)

README.md
hjir: I think the first link is not working for me, is that possible? But I like the idea of having three separate links for also qiskit and pennylane.

I'm not sure what you mean by the apex remark?

oschusler: The SSL certificate was not yet provided (apparently). I meant the root path (`/`). Apex is normally used for a hostname without subdomain (quantum-inspire.com for example). My mistake

README.md
oschusler: I think that makes sense, yes

hjir: I removed the dependencies and the note in the README file

README.md
hjir: Thanks, it works now (I updated the links)! Ah and I see what you mean.. I think quantum-inspire.com is good for now? Don't really see any other obvious choice

oschusler (approved):

PR #448, opened 2026-05-05
[QI2-1852] Set default number of shots if not shots available

oschusler:


quantuminspire/managers/resource_manager.py
```diff
@@ -179,15 +179,29 @@ def run_job_submission_flow(self, job_options: JobOptions, owner_id: int) -> Job
         Returns:
             The updated JobOptions with all remote resource IDs populated, including the submitted job ID.
         """
+        backend_type = self.get_backend_type(job_options.backend_type_id)
         commit = self._run_create_commit_flow(resource_options=job_options, owner_id=owner_id)
         file = self._run_create_file_flow(resource_options=job_options, commit_id=commit.id)
 
         batch_job: BatchJob = self._create_batch_job(backend_type_id=job_options.backend_type_id)
 
+        print(job_options.number_of_shots)
```
nienkeduetz: Is this print statement still necessary? Since we are printing this in the statement below.

oschusler: Valid point. Was a left-over from debugging

nienkeduetz (approved):

PR #449, opened 2026-05-05
[QI2-1853] filter projects

nienkeduetz:


oschusler:
@nienkeduetz I think you created a branch from master, not from feature/QI2-1149/rebuild-qi-cloud

quantuminspire/managers/resource_manager.py
```diff
@@ -386,21 +386,29 @@ def read_project(self, project_id: int) -> Project | None:
...
 
         Returns:
             A list of Project objects.
         """
+        filters = {}
+
+        if name and exact:
+            filters["name"] = name
+        elif name and not exact:
+            filters["search"] = name
```
oschusler: Is it cleaner to set the field based on the flag provided, instead of managing the assignment?


```suggestion
        if name:
            filter_field = "name" if exact else "search"
            filters[filter_field] = name
```

quantuminspire/cli.py
```diff
@@ -93,31 +93,50 @@ def initialize_project(
 
 
 @projects_app.command("list")
-def list_projects() -> None:
+def list_projects(
+    name: Optional[str] = typer.Option(None, help="Filter projects matching this name or description pattern"),
+    exact: bool = typer.Option(False, help="If set, match project name exactly instead of substring search"),
+    ids: bool = typer.Option(False, help="Only print the IDs of the projects."),
```
oschusler: More in line with other CLI's if you make this `-q/--quiet`

quantuminspire/cli.py
```diff
@@ -93,31 +93,50 @@ def initialize_project(
...
+        confirm_message = "Are you sure you want to continue?"
+
+    if not typer.confirm(confirm_message):
+        typer.echo("Aborted.")
+        raise typer.Abort()
+
     api = Api()
-    api.delete_projects()
-    typer.echo("All projects deleted successfully.")
+    api.delete_projects(project_ids, name, exact)
```
oschusler, line 138: Do we want to print the entire list of id's that have been deleted?

quantuminspire/managers/resource_manager.py
nienkeduetz: ✅

quantuminspire/cli.py
nienkeduetz: ✅

quantuminspire/cli.py
nienkeduetz: I think that's a good idea, I have added it to the API function, so they get printed too when the user uses the API directly.

oschusler (approved):

quantuminspire/util/connections.py
```diff
@@ -0,0 +1,33 @@
+from urllib.parse import urlparse
```
oschusler: I deleted this file in a previous commit. It should be gone here as well

tests/util/test_connections.py
```diff
@@ -0,0 +1,49 @@
+from unittest.mock import MagicMock
```
oschusler: I deleted this file in a previous commit. It should be gone here as well

oschusler (changes requested):

quantuminspire/util/connections.py
nienkeduetz: Removed

tests/util/test_connections.py
nienkeduetz: Removed

oschusler (approved):

PR #454, opened 2026-05-07
[QI2-1851] extend cli documentation

nienkeduetz:


quantuminspire/cli.py
```diff
@@ -18,16 +21,186 @@
...
+      qi config set backend_type_id 3 --algorithm my-algo
+    """
+
+
+@app.callback()
+def main() -> None:
+    """Quantum Inspire CLI.
+
+        With this Quantum Inspire CLI, you can manage your projects, algorithms and jobs on the Quantum Inspire system
+        from your command line. Before you start, please login to the system using:
```
oschusler: This is printed on my machine as 

> With this Quantum Inspire CLI, you can manage your projects, algorithms and jobs on the Quantum Inspire system     from your command line. Before you start, please login to the system using:

Note the non-breaking white space in between

oschusler (commented):
Very thorough documentation and I think it is good to also have it in the code. However, shouldn't we also have it in the markdown file?

quantuminspire/cli.py
nienkeduetz: Done!

oschusler (approved):

PR #459, opened 2026-05-09
[QI2-1869] Add queueing formation to CLI

oschusler:


quantuminspire/cli.py
```diff
@@ -234,6 +234,18 @@ def get_backend_type(
     console.print(backend_type.model_dump())
 
 
+@backend_types_app.command("queue")
+def get_queue(
+    backend_type_id: int = typer.Argument(None, help="The ID of the backend type to retrieve."),
+    user_only: bool = typer.Option(False, help="Whether to only show the user's position in the queue."),
```
hjir: Is it the user's position or number of jobs in the queue? The api docs say "whether to consider only the user's jobs in the queue." and I feel like that could be interpreted as two different things?

hjir (changes requested):

quantuminspire/cli.py
oschusler: This is the number of jobs in the queue

hjir: Then I think the docs used in the other two places is a clearer way to put it:
```
user_only: Whether to consider only the user's jobs in the queue.
```
The way it is phrased here (`"Whether to only show the user's position in the queue."`), it makes me think a position in the queue will be returned rather than a number of jobs.

hjir (approved):
It's a tiny documentation thing, so I am happy to approve and feel free to resolve the previous thread however you see fit :)

hjir (approved):

PR #483, opened 2026-06-08
[QI2-1818] Implemented end to end tests

NischalQuTech:


quantuminspire/managers/auth_manager.py
```diff
@@ -61,7 +46,7 @@ def refresh_tokens(self, host: Url) -> None:
 
     def login_required(self, host: Url) -> bool:
         """Return True if a full login flow is required for the host."""
-        return not (self._is_authenticated(host) and self._is_refresh_token_valid(host))
```
NischalQuTech, line 64: the deleted part was essentially checking the same thing, so I removed it

quantuminspire/utils/authentication.py
```diff
@@ -139,14 +138,3 @@ def revoke(self) -> None:
             raise AuthorisationError(
                 f"Token revocation failed with status code: {response.status_code}\n {response.text}"
             )
-
-
-class Configuration(compute_api_client.Configuration):  # type: ignore[misc]
```
NischalQuTech, line 144: this is already defined in `qi2_shared` so I removed it

quantuminspire/api.py
```diff
@@ -365,7 +366,6 @@ def _check_project_id(self) -> int:
         assert isinstance(project_id, int)
         return project_id
 
-    @_refresh_auth_tokens
```
NischalQuTech, line 368: this decorator should have been on the public method itself

quantuminspire/api.py
```diff
@@ -49,8 +50,6 @@ def wrapped(self: "Api", *args: P.args, **kwargs: P.kwargs) -> R:
 
             if self._login_required(hostname):
                 self.login(hostname)
-            else:
-                self._auth_manager.refresh_tokens(hostname)
```
NischalQuTech, line 53: this is not required, refresh happens on each api call see [here](https://gitlab.com/qutech-sd/quantum-inspire-2/platform/-/blob/master/src/python/libs/compute-api-client/compute_api_client/api_client.py?ref_type=heads#L646)

quantuminspire/managers/auth_manager.py
hjir: but wasn't the `_is_refresh_token_valid(host)` part calling `refresh_tokens` (aka also doing something and not just checking the validity?

quantuminspire/api.py
hjir: do I understand it correctly that the `update_params_for_auth` from the compute api client is therefore equivalent to refreshing tokens?

NischalQuTech: So we are using custom [Configuration](https://gitlab.com/qutech-sd/quantum-inspire-2/platform/-/blob/master/src/python/libs/compute-api-client/qi2_shared/client.py?ref_type=heads) class and its `auth_settings` method refreshes the token

quantuminspire/managers/auth_manager.py
NischalQuTech: Right, it was 
- Refreshing token and storing them. Storing the refreshed access token is not required as it gets refreshed at each api call (as we discussed the other day)
- Raising an `AuthorizationError` if the tokens didn't exist which is already being done by `_is_authenticated`.

quantuminspire/api.py
hjir: I see now, fair

hjir (approved):
Looks good to me!

hjir (approved):

PR #510, opened 2026-07-31
[QI2-1899] Added job logs to cli options

NischalQuTech:


quantuminspire/api.py
```diff
@@ -475,20 +475,29 @@ def get_job(self, job_id: Optional[int] = None, algorithm_name: Optional[str] =
 
     @_refresh_auth_tokens
     def get_job_logs(
-        self, job_id: int, n_logs: Optional[int] = None, poll_interval: float = 5.0, timeout: float = 30.0
+        self,
+        job_id: Optional[int] = None,
+        algorithm_name: Optional[str] = None,
```
hjir, line 480: do I understand correctly that if neither `job_id` or `algorithm_name` is provided, it tries to see if there is only one algorithm? maybe it would be good to add a note about this edge-case scenario?

README.md
```diff
@@ -219,6 +222,14 @@ qi algorithms status <algorithm_name>
 
 Use `--wait` and `--timeout <seconds>` to wait for the job to complete.
 
+### Get algorithm job logs
+
+```bash
+qi algorithms logs <algorithm_name>
+```
+
+Use `--n-logs` to return early if you already know how many logs you expected, `--poll-interval <seconds>` to set the polling interval (default: 5s) and `--timeout <seconds>` to set the timeout (default: 30s).
```
hjir, line 231: How do you know how many logs you expect (or maybe, in which situation would you know)?

quantuminspire/api.py
NischalQuTech: indeed. I updated the comments

README.md
NischalQuTech: if you look at foe example `tests/data_files/hqca_circuit.py`, you would see some print statements which get logged down. Here the user knows how many he expects and thus he can return early rather than waiting and keep polling for default 30s

hjir: okay, thanks for explaining!

hjir (approved):

